PR #1, opened 2019-05-10
Add tests for name parsing

chee:
Also fixes a bug where it was giving up after hitting any inline markdown 😊


tests/name.test.js
```diff
@@ -0,0 +1,40 @@
...
+xtest('more than one h1 is an error', async () => {
+	const { errors } = await runbookmd.parseRunbookString(here`
+		# hello monkey
+
+		# also this
+	`);
+
+	expect(errors.length).toBe(1);
+});
+
```
wheresrhys: idea for additional test - h1 preceded by any other content is an error

tests/name.test.js
```diff
@@ -0,0 +1,40 @@
...
+test('an h1 is parsed as name', async () => {
+	const { data, errors } = await runbookmd.parseRunbookString(
+		'# hello monkey',
+	);
+	expect(errors.length).toBe(0);
+	expect(data).toHaveProperty('name', 'hello monkey');
+});
+
+test('inline markdown is stripped', async () => {
```
wheresrhys: Should this throw an error instead? I'm not sure. What's your thinking?

tests/name.test.js
```diff
@@ -0,0 +1,40 @@
...
+	const { errors } = await runbookmd.parseRunbookString(here`
+		# hello monkey
+
+		# also this
+	`);
+
+	expect(errors.length).toBe(1);
+});
+
+test('no h1 means no name', async () => {
```
wheresrhys: Should this error too?

chee: Yes, i think this should be an error.

I'm currently learning about appending new nodes to the tree, I'll make this and the currently pending test create `Problem`s instead.

tests/name.test.js
chee: I thought here that people might want to use links or emphasis in the property keys and that we should strip all that away. And that we should also behave the same for `name` as for property keys.

But maybe it's better to not allow that at all and respond later if anyone wants that feature, rather than trying to guess what a user wants.

tests/name.test.js
chee: yes, that's a good idea.

tests/name.test.js
wheresrhys: Yeah, I would favour enforcing simplicity here, otherwise runbook.html could contain less semantic information than runbook.md

chee: We have two options then, one of keeping it exactly as it is, i.e.: "# one *two* three" gets imported as 

```js
{
  name: "one *two* three"
}
```

or that it creates a `Problem` if there is more than one child to the h1

tests/name.test.js
chee: i'm going to add this one later, it's a little different than the other tests. i've put it on my todo list for the project 😊.

chee:
I'm going to merge this, because the issues are resolved and I think it's good.

Which reminds me, I should add testing in CI this week too. 😊

PR #2, opened 2019-05-13
WIP include serverless, lambda and webpack in preparation for ingest endpoint.

GeoffThorpeFT:
# Problem
We need to introduce an /ingest endpoint (and a graphical interface)

# Solution
Lets use a serverless lambda and introduce webpack!

# Details
Lots of files added to enable serverless and webpack and two skeleton files for the endpoint and interface.

# Note
The above did require the introduction of a top level /src directory into which the existing /lib directory has been moved.


GeoffThorpeFT:
image

GeoffThorpeFT:
image

PR #3, opened 2019-05-13
Add description tests

chee:
adds 2 tests:
1. to ensure that anything at the top level that's not a name, or in a property is part of the description
2. to ensure nothing else gets munched up as description


test/description.test.js
```diff
@@ -0,0 +1,26 @@
+const runbookmd = require('..');
```
GeoffThorpeFT: Not sure I understand what a require of .. does.

chee: i'm requiring the main module, requiring the directory above :)

GeoffThorpeFT: Ah, so picks up index.js which itself exports runbook md. :)

PR #7, opened 2019-05-16
Add property name tests

chee:
Some changes made when writing the tests for name and description broke properties altogether, so this PR also contains fixes for those.
It also contains the commit from this: https://github.com/Financial-Times/runbook.md/pull/6, that'll clean up when #6 is merged.

I'd much rather be mocking the biz-ops schema library here, so that i'm not testing against for-real keys and testing that not-real keys don't exist. But from a discussion with @GeoffThorpeFT i understand that's something for a later day.


test/property-name.test.js
```diff
@@ -0,0 +1,27 @@
+const runbookmd = require('..');
+const here = require('outdent').default;
+
+test('a valid property name gets correctly coerced', async () => {
+	const { data } = await runbookmd.parseRunbookString(here`
+		# name
```
GeoffThorpeFT: Its probably a good idea here not to use both # name as the name of the system and 'property name' as the detail of the test. It took me a couple of goes to realise the property named 'name' is of no interest to the test (but is needed for formatting) and that the property named 'Contains xxxxxx' is what is being checked to confirm it is a valid property in the schema.

PR #8, opened 2019-05-16
Add property value coercion tests

chee:
This PR contains tests for property value coercion. It also contains the commits from my other 2 recent PRs, so will be easier to read once those have merged.

As well as adding these tests it makes some other changes:

* reduce subdocument headings by 2 levels rather than 1
  - i think these should be valid documents regardless of where they are used. 
  - when outputting them in runbooks.in.ft.com we can bump the heading back up a level if we are rendering them in a section with its own h1
* rename a couple of functions and files
* create arrays if a property type has `hasMany` in the schema. 
  - the implementation is naïve and there is a long comment in the file explaining why i think that's fine for now
* Remove coercers for non-system property types
  - as i'm currently testing with Real Biz Ops Schema Data, primitive types that are not used in Biz Ops schema can't be tested
  - if the code isn't being used and can't be tested, i think it should be deleted?
* insert an error to the errors array if a date is invalid.


test/property-value.test.js
```diff
@@ -0,0 +1,153 @@
...
+	const { errors } = await runbookmd.parseRunbookString(here`
+		# name
+
+		## last review date
+
+		mario's birthday
+	`);
+
+	expect(errors).toHaveLength(1);
+});
```
GeoffThorpeFT: You could also do with a test for URLs.  The primary URL is expected to be a URL and appear as a link on biz-ops-admin. Currently I'm seeing <url> - i.e. extra brackets.

GeoffThorpeFT:
```parseResult: {
data: {
name: "name with other fields",
primaryURL: "<https://sos.in.ft.com>",
hostPlatform: "aws"
},
errors: [ ]
},```

chee: I've added an item to take care of this, and will file a new PR today.

PR #11, opened 2019-05-21
Formatted submit response

GeoffThorpeFT:
# Problem
The output from the required /ingest endpoint was hidden within the form handler and was poorly formatted and tested.

# Solution
Return JSON and a status from /ingest to allow render by form hander.

# Implementation
Move code from form handler into ingest api endpoint to ensure json return. Improve form handler result formatting and error presentation. Write some tests for the /ingest endpoint.


GeoffThorpeFT:
image

GeoffThorpeFT:
image

GeoffThorpeFT:
image

GeoffThorpeFT:
image

GeoffThorpeFT:
One outstanding question re the implementation is whether to allow a Biz Ops update if the SOS validation fails (i.e. there are missing/incorrect fields).  I have assumed yes since its just a couple of missing fields.  I do NOT allow a Biz Ops update when the PARSE fails as that implies the format of the RUNBOOOK.MD as a whole is wrong (we could therefore end up with data in the wrong fields).

PR #12, opened 2019-05-22
Improve the scope, logging and viewing of the Biz Ops update.

GeoffThorpeFT:
# Problem
Forgot to allow for relationship updates and to provide the user with direct access to the updated Biz Ops record

# Solution
Improved call to Biz Ops and improved message on Biz Ops success

# Implementation
Pass relationshipAction parameter to Biz Ops Call.
Use lambda logger to log failure in Biz Ops call.
Add Biz Ops link to the o-message that is displayed on success after biz ops write.

### DIFF
--- a/src/ingestHandler.js
+++ b/src/ingestHandler.js
@@ -52,7 +52,7 @@ const ingest = async (username, userRequest) => {
 	if (!userRequest.writeToBizOps || userRequest.writeToBizOps === false) {
 		return success({
 			message:
-				'Parse & Validation Complete; Biz Ops Was NOT Updated at your request',
+				'Parse & Validation Complete. Biz Ops Was NOT Updated as you did not enable the writeToBizOps flag.',
 			details: ingestedDetails(parseResult, validationResult),
 		});
 	}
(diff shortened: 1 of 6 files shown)


src/ingestHandler.js
```diff
@@ -31,33 +34,34 @@ const ingestedDetails = (parseResult, validationResult, writeResult) => ({
...
 
 	const { json: validationResult } = await validate(parseResult.data);
 	if (!userRequest.writeToBizOps || userRequest.writeToBizOps === false) {
-		return success(
-			'Parse & Validation Complete; Biz Ops Was NOT Updated at your request',
-			ingestedDetails(parseResult, validationResult),
-		);
+		return success({
+			message:
+				'Parse & Validation Complete; Biz Ops Was NOT Updated at your request',
```
wheresrhys: perhaps make `at your request` more explicit e.g. `because you did not check the write to biz ops checkbox`

GeoffThorpeFT: Message has been improved.

PR #13, opened 2019-05-22
Parsing codes

GeoffThorpeFT:
# Problem
Codes are not treated as a special case during parsing which allows invalid values (codes that do not exists or are of the wrong type) and incorrect structure (supplying many values to a 1-2-1 relationship and vice versa).

# Solution
Parse codes into a known format that can be used to validate them against the schema and Biz Ops

# Implementation
Create a new checkBizOpsCodes function to check each code is of the right type and valid. Errors are produced in the 'parse' format and added as parse failures.


GeoffThorpeFT:
This is still WIP as the check for 'isNested' in the corercer really ought to be a check of isRelationship and isMany.  That would then enable relationship parsing to error if a single value was sent to an isMany or many values were sent to a !isMany.

src/lib/property-coercers/index.js
```diff
@@ -31,6 +31,26 @@ function split(subdocument) {
   `false` then `value` will be the problem message
 */
 module.exports = {
+	/*
+	  Huge assumption being made here that the only thing that will ever care
+	  about `hasMany` is `String` and `Subdocument` types, which in our case
+	  includes Codes.
+	*/
```
chee: This comment is a duplicate of the one below.

src/lib/property-coercers/index.js
```diff
@@ -31,6 +31,26 @@ function split(subdocument) {
...
+	Code(subdocument, { hasMany = false } = {}) {
+		if (hasMany) {
```
chee: quite a lot of repetition happening here now with the code from the two functions below, should probably be pulled out into a function

src/lib/property-coercers/index.js
```diff
@@ -31,6 +31,26 @@ function split(subdocument) {
...
+			return {
+				valid: true,
+				value: split(subdocument).map(code => ({
+					code: flattenNodeToPlainString(code),
+				})),
+			};
+		}
+		return {
+			valid: true,
+			value: { code: flattenNodeToPlainString(subdocument) },
```
chee: when we discussed this last SOS was going to be using the fact the nested fields were strings instead of objects as the indicator that this field was not a fleshed-out nested property. has this changed?

GeoffThorpeFT:
@chee Sorry for any confusion. @wheresrhys and I had a five minute chat about where the 'is this code valid?' check could be made. We concluded that runbook.md could take on the role since it is its responsibility to provide valid code data to the sos validation (and it is already making calls to biz ops).
The idea is therefore to call readBizOps() for each code (of any type) supplied by the user once the existing parse has confirmed the structure/content is valid.

GeoffThorpeFT:
However, the existing parse does not appear to correctly distinguish between hasMany and !hasMany in its parsed response.
```## dataOwner (which is not a hasMany)
geoff.thorpe
rhys.evans```
returns:
```{"code":"geoff.thorpe\nrhys.evans"}```

```## dependencies (which is a hasMany)
system-one
system-two
system-three```

returns:
```[{"code":"system-one\nsystem-two\nsystem-three"}]```

The second example does have the array syntax but has not split the parsed data into individual values.

**Comment by GeoffThorpeFT** [comment]
Oops, silly me. I forgot to prefix the individual dependencies with a 'list' indicator.  Once I did that then, yes, dependencies are returned as an array.
```
## dependencies
- system-one
-  system-two
- system-three
```
Returns:
```
[{"code":"system-one"},{"code":"system-two"},{"code":"system-three"}]
```

File: src/lib/property-coercers/index.js (hunk 1 above)
Comment (GeoffThorpeFT) [reply src/lib/property-coercers/index.js]: fixed - or at least improved :)

**Comment by GeoffThorpeFT** [comment]
Example error output when invalid codes are used:
image

**Comment by GeoffThorpeFT** [comment]
image

File: src/lib/property-coercers/index.js (hunk 2 above)
Comment (GeoffThorpeFT) [reply src/lib/property-coercers/index.js]: obsolete - replaced by your PR

File: src/lib/property-coercers/index.js (hunk 3 above)
Comment (GeoffThorpeFT) [reply src/lib/property-coercers/index.js]: Obsolete - replaced by your PR

File: src/ingest/code-validation.js [hunk 4]
```diff
@@ -0,0 +1,91 @@
...
+	const bizOpsErrors = Object.entries(bizOpsResponse.data)
+		.filter(([, value]) => value === null)
+		.map(([key]) => {
+			const [type, code] = key.split(/_(.+)/);
+			return `${code} is not a valid ${type} code`;
+		});
+	return { bizOpsData, errors: bizOpsErrors };
+};
+
+const checkBizOpsCodes = async (username, data) => {
```
wheresrhys: more descriptive name for this function?

src/ingest/code-validation.js
```diff
@@ -0,0 +1,91 @@
...
+	await schema.refresh();
```
wheresrhys: where is this function called from? If it's running within a lambda then the lambda wrapper should already be doing this (if not, then move this into there). schema refresh is supposed to be an application level task, rather than something for each individual task to have to do. Was it causing problems to not do it here?

src/ingest/code-validation.js
```diff
@@ -0,0 +1,91 @@
+const schema = require('../lib/get-configured-schema.js');
+const { queryBizOps } = require('../ingest/external-apis');
+
+const getRelationships = (systemSchema, data) =>
+	Object.assign(
```
wheresrhys: A bit more usual to use reduce to squish an array of entries into an object e.g. https://github.com/Financial-Times/biz-ops-api/blob/master/server/data/data-conversion.js#L53

src/ingest/code-validation.js
```diff
@@ -0,0 +1,91 @@
...
+			return `${code} is not a valid ${type} code`;
+		});
+	return { bizOpsData, errors: bizOpsErrors };
+};
+
+const checkBizOpsCodes = async (username, data) => {
+	await schema.refresh();
+	const systemSchema = schema.getTypes().find(type => type.name === 'System');
+	const relationships = getRelationships(systemSchema, data);
+	const bizOpsCodes = getTypesAndCodesFromRelationships(relationships);
```
wheresrhys: you're getting the properties that are relationships, which involves Object entries, then conversion back into an object. Then this function converts it back to entries and returns an array again.

Would be more efficient to just have a single function which does it all, and then you can avoid the converting back to an object in the middle

src/ingest/code-validation.js
```diff
@@ -0,0 +1,91 @@
...
+		...Object.entries(relationships).map(([, { type, code }]) =>
+			typeof code === 'string'
+				? [{ type, code }]
+				: code.map(value => ({ type, code: value })),
+		),
+	);
+
+const getUniqueCodes = codes =>
+	codes.filter(
+		(element, index, array) =>
```
wheresrhys: name element and object something more descriptive e.g. current and first

src/ingest/code-validation.js
```diff
@@ -0,0 +1,91 @@
...
+		});
+	return { bizOpsData, errors: bizOpsErrors };
+};
+
+const checkBizOpsCodes = async (username, data) => {
+	await schema.refresh();
+	const systemSchema = schema.getTypes().find(type => type.name === 'System');
+	const relationships = getRelationships(systemSchema, data);
+	const bizOpsCodes = getTypesAndCodesFromRelationships(relationships);
+	const uniqueBizOpsCodes = getUniqueCodes(bizOpsCodes);
```
wheresrhys: for something like this I'd consider lodash.uniqBy. I'm not a _huge_ fan of lodash as it has such a large api and a lot of the functions are not that intuitively named or signatured, but uniqBy is a good one to know; for larger lists it will be way more efficient than any simple attempt to implement uniqueness

src/ingest/code-validation.js
```diff
@@ -0,0 +1,91 @@
...
+	codes.filter(
+		(element, index, array) =>
+			array.findIndex(
+				object =>
+					object.type === element.type &&
+					object.code === element.code,
+			) === index,
+	);
+
+const formatGraphQLQuery = bizOpsCodes =>
```
wheresrhys: rename to buildGraphQLQuery

src/ingest/code-validation.js
```diff
@@ -0,0 +1,91 @@
...
+			.map(([key, value]) => {
+				const code = key.split(/_(.+)/)[1];
+				return { [code]: value };
+			}),
+	);
+	const bizOpsErrors = Object.entries(bizOpsResponse.data)
+		.filter(([, value]) => value === null)
+		.map(([key]) => {
+			const [type, code] = key.split(/_(.+)/);
+			return `${code} is not a valid ${type} code`;
```
wheresrhys:
```suggestion
			return `There is no ${type} with code ${code}stored in Biz Ops`;
```

src/ingest/code-validation.js
GeoffThorpeFT: Now named validateCodesAgainstBizOps

src/ingest/code-validation.js
GeoffThorpeFT: Removed  - it was already being called.

src/ingest/code-validation.js
GeoffThorpeFT: Its now a simplified single function.

src/ingest/code-validation.js
GeoffThorpeFT: done

src/ingest/code-validation.js
GeoffThorpeFT: done

src/ingest/code-validation.js
GeoffThorpeFT: done

src/ingest/code-validation.js
GeoffThorpeFT: Merged into another function

src/ingest/code-validation.js
```diff
@@ -0,0 +1,82 @@
...
+			.map(([property, value]) => ({
+				type: systemSchema.properties[property].type,
+				code: value,
+			}))
+			.map(({ type, code }) =>
+				typeof code === 'string'
+					? [{ type, code }]
+					: code.map(value => ({ type, code: value })),
+			),
+	);
```
doramatadora: I like this chain of functions for readability, however I think we should slash complexity here to a single pass by using [`Array.reduce`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/reduce):
```js
const getTypesAndCodesFromRelationships = (systemSchema, data) =>
	Object.entries(data).reduce((accumulator, [property, value]) => {
		const { type, isRelationship } = systemSchema.properties[property];
		if (isRelationship) {
			if (typeof value === 'string') {
				accumulator.push({ type, code: value });
			} else {
				accumulator.push(...value.map(code => ({ type, code })));
			}
		}
		return accumulator;
	}, []);
```

src/ingest/code-validation.js
```diff
@@ -0,0 +1,82 @@
...
+
+const getUniqueCodes = codes =>
+	codes.filter(
```
doramatadora: Agreed with @wheresrhys here – a good use case for lodash

src/ingest/code-validation.js
```diff
@@ -0,0 +1,82 @@
...
+			({ type, code }) =>
+				`${type}_${code.replace(
+					/[-.]/g,
+					'_',
+				)}:${type} (code:"${code}") {code,name}`,
+		)
+		.join('\n')}
+	}`;
+
+const formatBizOpsResponse = bizOpsResponse => {
```
doramatadora: Just because we are in Lambda-land I would reduce complexity with a single iteration., e.g.:
```js
const bizOpsData = {};
const errors = [];
Object.entries(bizOpsResponse.data).map(([key, value]) => {
	const [, code] = key.split(/_(.+)/);
	if (value === null) {
		errors.push(
			`There is no ${type} with a code of ${code} stored within Biz Ops`,
		);
	} else {
		bizOpsData[code] = value;
	}
});
return { bizOpsData, errors };
```

src/ingest/code-validation.js
GeoffThorpeFT: Obsolete - replaced by refactored code.

src/ingest/code-validation.js
GeoffThorpeFT: Thanks. Much better than my code.

src/ingest/code-validation.js
GeoffThorpeFT: Thanks. Much better than my code.

src/ingest/code-validation.js
GeoffThorpeFT: Obsolete - now done as part of another function

PR #14, opened 2019-05-22
correctly emit enum values when set

chee:
Had been reading an empty value off a string 😊


chee:
https://trello.com/c/RXCWXQ8H/656-ensure-runbookmd-parses-and-returns-values-from-enums-eg-servicetier-and-lifecyclestage

test/property-value.test.js
```diff
@@ -60,17 +60,42 @@ test('string types are coerced to string', async () => {
...
+	expect(data.serviceTier).toBe('Platinum');
+});
+
+test('enums types with incorrect values produce error', async () => {
+	const { data, errors } = await runbookmd.parseRunbookString(here`
+		# name
+
+		## Service tier
+
+		Cardboard
```
wheresrhys: :cheese:

PR #15, opened 2019-05-24
Sample runbook

GeoffThorpeFT:
# What
A fully populated example of the critical text within a RUNBOOK.MD file.

# Why
So we can offer the content as a sample within the GUI and offer a link to the raw and formatted text within this repo.

# How
The creation of an EXAMPLE.MD file

# Caveat
Not a perfect runbook; just an example of how to populate all the fields.


EXAMPLE.MD
```diff
@@ -0,0 +1,132 @@
+# This is the name of the system
```
wheresrhys: please rename to EXAMPLE.md (this involves a faff of renaming the file to something completely different, committing, then renaming to EXAMPLE.md). This is _such_ a nitpick I know, so I'm sorry, but inconsistent casing can sometimes lead to hours lost, tearing out hair dealing with mystery errors, so best avoided (the only reason uppercase for the file name makes sense is because github creates README.md by default, so establishes the convention)

src/templates/components/input-fields.jsx
```diff
@@ -64,10 +64,10 @@ const ApiKey = ({ bizOpsApiKey }) => (
 	</div>
 );
 
-const RunbookMd = ({ content }) => (
+const RunbookMd = ({ exampleContent, content }) => (
 	<div className="validation-content">
 		<label className="o-forms__label" htmlFor="content">
-			The RUNBOOK.MD Content
+			The RUNBOOK.MD Content.
```
wheresrhys:
```suggestion
			The RUNBOOK.md Content.
```

EXAMPLE.MD
```diff
@@ -0,0 +1,132 @@
...
+## hostPlatform
+AWS
+
+## deliveredBy
+delivery-team
+
+## supportedBy
+support-team
+
+## knownAboutBy
```
chee: I think we should use the title case format in this example file, so

```suggestion
## knownAboutBy
## Known about by
```
etc.

because that's what I'd hope we'll be recommending people to do, and what I believe people want

PR #18, opened 2019-05-29
Validate coerced properties with biz-ops-schema

chee:
Ah, so after doing the work to make it run the `validateProperty` on fields after they’ve been produced by the coercer, i went to write a test for it.
The only thing that uses a string pattern in `System` so far is `Code` (which we aren’t going to allow them to set in the runbook, haha).
I’ve pushed the code up anyway, because it will work once we add something like a url regex to the validation strings.


src/lib/runbook.md.js
```diff
@@ -21,11 +19,12 @@ async function runbookMd() {
 
 	const typeNames = new Set(types.map(type => type.name));
 
+	const validateProperty = (key, value) => {
+		return schema.validateProperty('System', key, value)
```
doramatadora: Prettier will complain about a missing semicolon here and on the line below

src/lib/runbook.md.js
```diff
@@ -38,6 +37,9 @@ async function runbookMd() {
 			primitiveTypesMap: schema.primitiveTypesMap,
 			enums: schema.getEnums(),
 		})
+		.use(validateBizopsProperties, {
+			validateProperty
```
doramatadora: Prettier will complain about a missing comma here

doramatadora:
Looks great 👍 Requested changes are linting.

src/lib/runbook.md.js
chee: ah, thanks 😊 i'll do a little rebase

chee:
> Looks great 👍 Requested changes are linting.

Thanks !! i did a little history rewriting and now the linting has always have been right 😊

PR #21, opened 2019-06-03
Return error object rather than error message

chee:
So that the `line` can be used for other things some day


src/templates/components/output-fields.jsx
```diff
@@ -103,12 +103,16 @@ const ParseErrors = ({ errors }) => (
...
+					</th>
 				</tr>
 			</thead>
 			<tbody>
-				{errors.map(message => (
+				{errors.map(error => (
 					<tr>
-						<td>{message}</td>
+						<td>{error.message}</td>
+						<td>{error.line}</td>
```
GeoffThorpeFT: I would usually swap the error line and error message so that the line was shown first.
We should then consider updating the style of the area text box so we can show line numbers.

GeoffThorpeFT: Ooh, looks like someone has already thought of that: https://stackoverflow.com/questions/52369434/textarea-with-alternate-rows-and-line-numbers

chee: this is a great idea. we could use marijn haverbeke's great [codemirror](https://codemirror.net/) so they get syntax highlighting as well as line numbers (and even multiple cursor support!)

chee: for now i'm going to make this print the exact same message as it did before

PR #23, opened 2019-06-13
Convert to a monorepo, first pass

chee:
## Why?

we have some closely related projects in this repo, but that are distinct from one another. A monorepo is a tool we have for that until we write a new source control system.

## Some issues:

* ~the Makefile in the root knows too much about the Makefile in the injester~ `Fixed`!
* ran into some issues with the autodetects in `eslint-config-reliability-engineering` that i currently think i have to work around by having the root package depend on things that it doesn't use (to make `usesJsx` and `usesJest` be true)


chee:
> Nice, clear structure ..... but I would like the subdirectory named ingester not injester. [en.wiktionary.org/wiki/ingest](https://en.wiktionary.org/wiki/ingest)
> [urbandictionary.com/define.php?term=injest](https://www.urbandictionary.com/define.php?term=injest)

oh, haha !!

Makefile
```diff
@@ -27,40 +27,16 @@ PROJECT_NAME=biz-ops-runbook-md
...
-		then webpack-dev-server --mode development; \
-		else webpack --mode production; \
-	fi
-
-run: clean
-	@concurrently "make build-statics" "make transpile" "make serverless-offline"
+	npx athloi run test
+
+ingester:
+	$(MAKE) -C lambdas/ingester $(MAKECMDGOALS)
```
wheresrhys: I have never seen such a thing before. This is an out of the box make thing?

Makefile
```diff
@@ -27,40 +27,16 @@ PROJECT_NAME=biz-ops-runbook-md
...
+
+serverless-offline: ingester
```
wheresrhys: Is ingester all the existing lambdas for the form?

Would it perhaps make more sense to have lambdas/ui & lambdas/api (or maybe lambdas/machine to capture the sense of 'all the lambdas that are api, or stream, or cron or ... driven)... assuming that one goal of this work is to reduce wasteful running of UI build steps on non UI lambdas

wheresrhys: or, mmm, I don't know. Just a thought

lambdas/ingester/src/ingestHandler.js
```diff
@@ -1,5 +1,5 @@
 const logger = require('@financial-times/lambda-logger');
-const runbookMd = require('./lib/runbook.md');
+const runbookMd = require('@financial-times/runbook.md-parser');
```
wheresrhys: How does this work in local dev?

libraries/parser/.eslintrc.js
```diff
@@ -0,0 +1,5 @@
+'use strict';
+
+module.exports = {
```
wheresrhys: does each subproject need its won eslintrc? Can eslint not be configured to look 2 steps up?

Is it not an issue with lintspaces (which relies on .editorconfig) and other tooling such as style-lint?

package.json
```diff
@@ -1,80 +1,33 @@
...
+  "name": "runbooks.md",
   "version": "1.0.0",
   "description": "Automated import of runbook.md",
   "main": "index.js",
-  "directories": {
-    "test": "tests"
-  },
   "scripts": {
-    "test": "jest"
+    "test": "athloi run test"
```
wheresrhys: prefer not to define test scripts in package.json - creates confusion as makefile is the canonical task runner. The presence of jest here before is a mistake?

wheresrhys:
> the Makefile in the root knows too much about the Makefile in the injester
 can L32 - EOF be dropped in favour of a normal make include???

```
-include lambas/ingester/Makefile
```

libraries/parser/.eslintrc.js
wheresrhys: ah just seen your comment in the PR description. Maybe t's fixable in rel-engage. What do you think @limess?

Limess: That package currently looks up one package.json. It could be configured to look further up for shared dependencies, I'm not sure if there's an existing convention here with monorepos and dependency detection we can use?

Limess:
I do wonder if npm scripts would simplify this a lot and make it a lot more understandable given the monorepo structure. Generally tools like lerna/athloi support running nested scripts or shared scripts from the top level as there out-of-the-box scripting mechanism.

Makefile
Limess: Apparently, I've also never seen it: https://www.gnu.org/software/make/manual/html_node/Goals.html

Makefile
Limess: I don't understand what this is doing at a glance.

lambdas/ingester/src/ingestHandler.js
Limess: I assumed `athloi` symlinked everything together like `lerna` but that doesn't seem to be the case, I'd also be interested to know this!

libraries/parser/.eslintrc.js
chee: yes, this is a problem specific to our rootPackageJson feature detection. there's definitely a way to solve that in rel-engage, but for now i've done this.

A comment should be added in each of these sub-eslints to explain that, i'll put it on my list

Makefile
chee: yeah, i would love a better way of doing this, but i have nothing yet

Makefile
chee: yes, absolutely it would. though, i think that work is something dora was planning on tackling after athloi was set up.

@Limess which? i'll add a comment explaining what it's doing and why, until we have a better system set up for running the subtasks (i don't really know serverless and was hoping to be able to make these initial changes without learning serverless)

package.json
chee: ah, yes it must be. i think it was `"test": "jest"` before i added rel-engage

lambdas/ingester/src/ingestHandler.js
chee: in the package.json this is

```json
"@financial-times/runbook.md-parser": "../../libraries/parser"
```

and running `athloi version 1.0.0` will update the versions in all the package.jsons to point at the correct version once published.
then `athloi publish` will publish them in their topologically sorted order :)

we can add steps for this to the `Makefile`?

package.json
chee: we should also set up a jest config in the root to be the thing that runs the tests for everyone

chee:
> > the Makefile in the root knows too much about the Makefile in the injester
> 
> can L32 - EOF be dropped in favour of a normal make include???
> 
> ```
> -include lambas/ingester/Makefile
> ```

unfortunately not, because then it won't run in the correct directory.
we need make to run in the package directory.
there is very likely a better way of organising the serverless and tasks, but this first pass was meant to behave identically to the previous organisation and iterate from there.

chee:
> I do wonder if npm scripts would simplify this a lot and make it a lot more understandable given the monorepo structure. Generally tools like lerna/athloi support running nested scripts or shared scripts from the top level as there out-of-the-box scripting mechanism.

maybe!

`make` support this too, with `make -C`

Makefile
Limess: Any of the above tasks

Thinking about it now I assume they're running the target in the context of the lambda's makefile?

Limess: As an aside I think all these need a `.PHONY` target defining?

.circleci/config.yml
```diff
@@ -1,153 +1,153 @@
...
+            keys:
+                - *npm_cache_key
+                - v1-dependency-npm-
+
+    cache_node_modules: &cache_node_modules
+        save_cache:
+            key: *npm_cache_key
+            paths:
+                - ./node_modules/
+                - ./lambdas/injester/node_modules/
```
doramatadora: *`ingester`

.circleci/config.yml
```diff
@@ -1,153 +1,153 @@
...
-            NODE_ENV: production
+    install:
+        executor: node
+        steps:
+            - checkout
+            - attach_workspace_at_project
+            - *restore_node_modules
+            - run:
+                  name: Install dependencies
+                  command: npm install --no-save
```
Limess: Is this still correct for the all the monorepo subpackages? Does the cache still apply?

lambdas/ingester/README.md
```diff
@@ -0,0 +1,12 @@
+# Automated Import of RUNBOOK.md files
+
+This utility offers the user the opportunity to create and upload a RUNBOOK.md
+file into Biz Ops to populate all the critical fields.
+
+In this **BETA** implementation we offer both a GUI and a /ingest endpoint (to
+which you can POST RUNBOOK.md) content
+
+A sample of the format/content of the RUNBOOK.md file can be seen
+[here](./EXAMPLE.md)
```
doramatadora: Please update the reference to this

lambdas/ingester/src/ingestHandler.js
doramatadora: Add steps to `circle.yml`

libraries/parser/.eslintrc.js
doramatadora: Can we write up a ticket for this? In the meantime a comment in our code would be good

.circleci/config.yml
chee: ah, i think i've told it to cache the other things but i haven't told it to build them

chee:
> I do wonder if npm scripts would simplify this a lot and make it a lot more understandable given the monorepo structure. Generally tools like lerna/athloi support running nested scripts or shared scripts from the top level as there out-of-the-box scripting mechanism.

leaning towards this more and more

Makefile
chee: I have something better now. the circle ci config is now using `make -C` instead

Makefile
chee: i've removed all this now ^_^

.circleci/config.yml
chee: it now builds them

PR #24, opened 2019-06-13
Add lambda to serverless configuration

OttlikG:
With this change, I'm expecting to get the different services and policies linked together so I can test out my lambda.


OttlikG:
Yes, it has we just pushed this bit so @nl-ria can have a look at the configuration.

src/releaseHandler.js
```diff
@@ -0,0 +1,221 @@
...
+	childLogger.info({
+		event: 'RELEASE_TRIGGERED',
+		value: event,
+	});
+
+	const parsedRecords = event.Records.map(parseRecord(childLogger))
+		.filter(productionChanges)
+		.filter(Boolean);
+
+	const runbookMDs = await fetchRunbookMds(parsedRecords, childLogger);
```
wheresrhys: I got confused a little reading this. I was thinking "why need multiple runbook.md's for one repo?"... I forgot that each event can contain several change reocrds for different systems. IMHO it'd be clearer to 
1. get systems/repositories list from records
2. then iterate over that list with a single `processRunbookMd` function

src/releaseHandler.js
```diff
@@ -0,0 +1,221 @@
...
+				}
+			},
+		),
+	);
+
+	return resolved;
+};
+
+const productionChanges = (payload = {}) =>
+	['p', 'prod', 'production'].includes(payload.environment);
```
wheresrhys: this logic should - and maybe already does - live in change api. It's not runbook.md's job to define which strings map to `isProduction: true`, and risks falling out of step.

wheresrhys: If it doesn't exist in change-api/the particular change-api stream this listens to, then can you add a tech debt ticket to address this if it's not trivial to fix properly up front

src/releaseHandler.js
```diff
@@ -0,0 +1,221 @@
...
+						.replace('https://github.com/', '')
+						.match(/[\w-]+\/[\w-]+/)[0];
+					let runbookContent;
+
+					runbookContent = await getRunbookFromCommit(
+						commit,
+						gitRepositoryName,
+					);
+
+					if (!runbookContent) {
```
wheresrhys: what if it's a PR where the first commit updates runbook.md, and then some other commits also update runbook.md. Will this risk only applying the changes in the first commit?

If not, it could do with some comments to explain why this logic works

serverless.yml
```diff
@@ -59,6 +64,17 @@ functions:
       - http:
           path: /ingest
           method: post
+  releaseHandler:
```
wheresrhys: rename to `releaseLogHandler` for clarity?

src/releaseHandler.js
OttlikG: The integration we suggest with change-api use `CIRCLE_SHA1` which is the last commit in the build. If the last commit contains `RUNBOOK.md` then we use that if not we look up the PR based on the commit and then find if there is a change in the descriptor file if so we will ingest it.

serverless.yml
OttlikG: That's a good shout, thanks

src/releaseHandler.js
OttlikG: Yes, it does thanks for highlighting.

src/releaseHandler.js
OttlikG: The reason why I split this into different functions because the function names document itself.
I can create a function called `processRunbookMd` that wraps all logic and handle the requests in a Promise.all so we keep parallelisation.

Makefile
```diff
@@ -62,5 +62,9 @@ build-statics:
 		else webpack --mode production; \
 	fi
 
+export IS_RUNNING=$(shell docker inspect --format '{{.State.Running}}' localstreams)
 run: clean
-	@concurrently "make build-statics" "make transpile" "make serverless-offline"
+	@if [ "$$IS_RUNNING" != true ]; then \
+		docker run -d --name localstreams -p 4567:4567 instructure/kinesalite ; \
```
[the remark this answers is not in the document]
OttlikG: Yes that was there just need to add that bit back to the Makefile.

OttlikG:
Thanks for the suggestions. I'll address soon.
I also opened up a new PR because this merge got a lot of conflict with the new monorepo structure.
So the history of commits may change. Will try to pair with @chee on this.
https://github.com/Financial-Times/runbook.md/pull/28

Makefile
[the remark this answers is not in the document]
OttlikG: I was waiting for the extra policy being merged so I can conditionally add stream-creation if not exist.

OttlikG:
Closing this in favour of this repo: https://github.com/Financial-Times/runbook.md/pull/46

PR #28, opened 2019-06-20
Merge release lambda

OttlikG:
With this change, I'm expecting to get the different services and policies linked together so I can test out my lambda.

I created this pr to test if everything is working after merging my changes into the monorepo.


wheresrhys:
I already left a few comments on https://github.com/Financial-Times/runbook.md/pull/24, which looks pretty much identical

wheresrhys:
I find it a bit easier to follow now - thanks.

Have we considered what all the failure states are and what the appropriate action to take is in each case? e.g. at the moment an error when fetching a runbook is caught and swallowed, while an error when writing a runbook to biz ops is thrown. I think we need to think carefully for each about whether we want to
- throw (which - for kinesis driven lambdas - means the function exits, but will keep retrying this batch forever), swallow the error or retry the api call. My gut feeling is 2-3 retries (to allow for api flakiness) followed by a swallow
- who to notify on failure. Should this function include a slack integration? (we already have a slack channel from the change api event)

wheresrhys:
Also needs tests - writing runbooks correctly to biz ops is pretty critical, so need to back it up with tests

OttlikG:
Yes, I can talk through with @doramatadora where to throw the errors. When I call `ingest` it will return 400 on invalid runbook.md content which won't throw. Propagating the error is more important when there is an internal error during ingest call.

I will definitely write test's just wanted to get the first round of feedback or refactoring. Just spare a little bit on typing and some time.

OttlikG:
I'm closing this branch in favor of this PR:
https://github.com/Financial-Times/runbook.md/pull/46

Merge conflicts due to the new monorepo structure give us a bit of a headache.
Some of our tools (secret squirrel, athloi) weren't prepared to handle monorepos in our use case so we have to migrate away from this solution in the future.

PR #30, opened 2019-06-24
Test that the error messages are a little good

chee:
add a couple tests to check the error messages mention the things the errors are about ^_^


libraries/parser/test/property-value.test.js
```diff
@@ -83,6 +84,9 @@ test('enums types with incorrect values produce error', async () => {
 
 	expect(data.serviceTier).toBe(undefined);
 	expect(errors.length).toBe(1);
+	const [{ message }] = errors;
+	expect(message).toMatch(/\bcardboard\b/);
+	expect(message).toContain('ServiceTier');
```
GeoffThorpeFT: Can we be more specific about the message?

chee: if we're going to be more specific we might as well use jest snapshots and test that the message doesn't change at all, which im fine with

PR #31, opened 2019-06-24
hack to make sure example content is included

wheresrhys:
# Why
webpack serverless is not providing the example.md file in prod

# What
Copies example.md into example.js


lambdas/ingester/src/formHandler.js
```diff
@@ -9,10 +9,12 @@ const { createLambda } = require('./lib/lambda');
...
-const exampleContent = readFile(
-	resolvePath(__dirname, '..', '..', '..', 'docs', 'example.md'),
-	'utf-8',
-);
+// const exampleContent = readFile(
+// 	resolvePath(__dirname, '..', '..', '..', 'docs', 'example.md'),
+// 	'utf-8',
+// );
+
+const exampleContent = require('../../../docs/example.md');
```
doramatadora: This should just be `example` in the ingester path, but if this `require` works for `example.md`, even better

wheresrhys: I had meant to leave it in the docs directory

lambdas/ingester/example.js
```diff
@@ -0,0 +1,177 @@
+module.exports=`# This is the name of the system
```
GeoffThorpeFT: Is this file used? Or are we just requiring the example.md from docs?

PR #38, opened 2019-06-25
Dora/fix deploy package

doramatadora:


lambdas/ingester/Makefile
```diff
@@ -22,7 +22,10 @@ clean:
 	rm -rf dist/
 
 transpile:
-	serverless webpack
+	@if [ -z $(CI) ]; \
```
GeoffThorpeFT: When does transpile get called?

PR #43, opened 2019-06-26
make the entry form behave correctly

doramatadora:
- fixes static builds (I accidentally reversed development and production while testing)
- fixes broken code validator (due to a named import in the wrong place)
- better client-side styling
- tech hygiene: code splitting and DRY


lambdas/ingester/src/templates/components/input-fields.jsx
```diff
@@ -1,66 +1,8 @@
...
-			aria-hidden="true"
-		>
-			{label}
-		</span>
-	</label>
-);
+const addLineNumbers = require('add-line-numbers');
+const { TextInput } = require('./text-input');
+const { RadioButton } = require('./radio-button');
+const { FormField } = require('./form-field');
```
GeoffThorpeFT: Why are these imports all destructured when there is only one function in each source file?  Should those source files not have explicit modules.export instead of exports.<function>

lambdas/ingester/src/templates/components/output-fields.jsx
```diff
@@ -1,6 +1,6 @@
 const { h } = require('hyperons');
 
-const Message = ({ status, message, linkText, linkUrl }) => (
+exports.Message = ({ status, message, linkText, linkUrl }) => (
```
GeoffThorpeFT: Why do we need to export each function individually?

lambdas/ingester/src/templates/components/output-fields.jsx
```diff
@@ -23,147 +23,104 @@ const Message = ({ status, message, linkText, linkUrl }) => (
...
+		]),
+	};
+	return (
+		<div className="validation-errors">
+			<Table {...tableProps} />
+		</div>
+	);
 };
+
+exports.Table = Table;
```
GeoffThorpeFT: Why export Table in a different way?

GeoffThorpeFT:
I really like the idea of line number visibility but perhaps that should be in its own PR.

lambdas/ingester/src/templates/components/output-fields.jsx
doramatadora: Easier to know what's exported/important without reaching the end of the file or fancy IDE plugins. This is all probably weird to see in CommonJS. But it's present in the code in other places.

lambdas/ingester/src/templates/components/input-fields.jsx
doramatadora: It's a matter of personal preference – explicit over implicit. The cognitive overhead of remembering what all the code does at the time of writing this PR was too much, but I would argue that named exports are always preferable to default: https://humanwhocodes.com/blog/2019/01/stop-using-default-exports-javascript-module/

lambdas/ingester/src/templates/components/output-fields.jsx
doramatadora: It is exported [in the same way](https://stackoverflow.com/questions/16383795/difference-between-module-exports-and-exports-in-the-commonjs-module-system), but having `exports.Table` in line with the assignment would make referencing in functions in the same file uglier. I will separate this into its own file as a shared helper

doramatadora:
Will separate line numbers into another PR

PR #47, opened 2019-06-26
Return nested info from Biz Ops during code validation

GeoffThorpeFT:
## What
Ensure the validation of runbook.md content passes all the nested fields to the sos/validate endpoint.
## Why
SOS validation fails with a long list of errors for each Group, Team, Person referenced in runbook.md since only the code is provided.
## How
During the validation of the codes (via the graphql call to Biz Ops) obtain all the nested fields (a different set for each type) and use then to create a fully populated runbook object (a simulation of the JSON that a runbook render would produce).


GeoffThorpeFT:
This currently has two quirks:
1) It makes use of a 'local' copy of the parser since references to @financial-times/runbook-md-parser we failing locally
2) conflicts with the recent change to the graphql query required by documentation-day - which escapes special characters.

lambdas/ingester/src/commands/ingest.js
```diff
@@ -1,6 +1,6 @@
 const runbookMd = require('../lib/parser');
 const { validate, updateBizOps } = require('../lib/external-apis');
-const { validateCodesAgainstBizOps } = require('../lib/code-validation');
+const { expandCodesIntoBizOpsData } = require('../lib/code-validation');
 
 const transformIngestedDetails = (
```
wheresrhys: could give this a more descriptive name - what's it transforming things for/to?

lambdas/ingester/src/commands/ingest.js
```diff
@@ -29,30 +29,26 @@ const ingest = async (username, payload) => {
...
-	const parseCheckResult = { ...parseResult, ...checkResult };
-	parseCheckResult.errors = [
-		...(parseResult.errors || []),
-		...(checkResult.errors || []),
-	];
 
-	if (parseCheckResult.errors.length) {
+	parseResult.errors.push(...expandedResult.errors);
+	if (parseResult.errors.length) {
 		throw decorateError({
```
wheresrhys: just a thought that throwing errors might not be the best way to handle this. In a sense, it's a still a success state, because we're successfully telling the user what is right/wrong with their runbook.md, as we intended to do. So return objects/arrays of error objects might be better. Maybe not for this PR though

lambdas/ingester/src/lib/code-validation.js
```diff
@@ -1,6 +1,16 @@
...
 
+const relatedBizOpsFields = {
+	Group: 'name,isActive',
+	Healthcheck: 'url,isLive',
+	Person: 'name,email,phone,isActive',
+	Repository: 'url,isArchived',
+	System: 'name,serviceTier,lifecycleStage',
+	Team:
+		'name,email,slack,phone,supportRota,contactPref,isActive,productOwners{code,name,email,phone,isActive},techLeads{code,name,email,phone,isActive},group{code,name,isActive}',
+};
```
wheresrhys: woudl this still work if you turned this into a multiline string for legibility?

lambdas/ingester/src/lib/code-validation.js
```diff
@@ -27,15 +37,18 @@ const getTypesAndCodesFromRelationships = (systemSchema, data) =>
 		return accumulator;
 	}, []);
 
+const replaceReservedCharacters = string => string.replace(/[-,]/g, '_');
```
wheresrhys: I think Dora found some more of these yesterday

lambdas/ingester/src/lib/code-validation.js
doramatadora: Or an array with `.join()` at the end, in case we need to add / remove properties easily.

lambdas/ingester/src/lib/code-validation.js
doramatadora: Yes, best to extract that into a sanitizer function. A `.replace('-','_').replace(/\W/g,'')` is necessary otherwise the graphQL queries break

lambdas/ingester/src/commands/ingest.js
GeoffThorpeFT: transformCodesIntoNestedData

lambdas/ingester/src/commands/ingest.js
GeoffThorpeFT: Ooh - For another PR

lambdas/ingester/src/lib/code-validation.js
GeoffThorpeFT: Done

lambdas/ingester/src/lib/code-validation.js
GeoffThorpeFT: Done

wheresrhys:
Last week I did this https://github.com/Financial-Times/runbook.md/commit/e6108d527b2f74387a6097dfaaca2cc910b5a3f8 to suppress confusing errors relating to teams which cannot be fixed in the runbook.md file. 

Now that we are including full data in order to generate SOS scores we need to make sure that we either
- continue to suppress these errors
- display them in a much reduced way which makes it clear that the place to fix them is not runbook.md (e.g. link to the team's page)

Suggest the first option is best for now

GeoffThorpeFT:
We have to be careful when suppressing errors with deliveredBy and supportedBy as they are of two kinds:
- the absence of a team in either field (which requires the team id to be provided in the system field)
- errors within team attributes (which require the team to be updated)
The current quick fix to avoid confusing errors is hiding both the above so runbook.md is not failing for missing delivery/support teams.

GeoffThorpeFT:
Code now checks the error properties to see if they are system related before allowing the associated error messages to be output.

PR #55, opened 2019-06-27
add github webhook listener and functionality

doramatadora:
# Why
Better UX: we want to automate the runbook validation process as much as possible to encourage a better quality of runbooks, organisation-wide.

# What
- A github webhook handler 
- Persistent validation results
- A status page which supports permalinks 

This enables the `runbook.md` GitHub app to:
1. Listen to push events
2. Ingest RUNBOOK.MD at that point in time
3. Store results
4. Post commit statuses


lambdas/ingester/serverless.yml
```diff
@@ -14,6 +14,7 @@ provider:
     timeout: 30
     deploymentBucket: artefacts.biz-ops-runbook-md.${env:AWS_ACCOUNT_ID}
     environment:
+        APP_ID: ${env:APP_ID}
```
wheresrhys: GITHUB_APP_ID?

lambdas/ingester/serverless.yml
```diff
@@ -25,8 +26,10 @@ provider:
         CONSOLE_LOG_LEVEL: info
         ENVIRONMENT: ${env:ENVIRONMENT}
         NODE_ENV: ${self:custom.${self:provider.stage}.NODE_ENV, 'production'}
+        PRIVATE_KEY_STRING: ${env:PRIVATE_KEY_STRING}
```
wheresrhys: GITHUB_PRIVATE_KEY_STRING?

lambdas/ingester/serverless.yml
```diff
@@ -25,8 +26,10 @@ provider:
...
         STAGE: ${self:provider.stage}
         SYSTEM_CODE: ${self:service}
+        WEBHOOK_SECRET: ${env:WEBHOOK_SECRET}
```
wheresrhys: GITHUB_WEBHOOK_SECRET?

lambdas/ingester/src/commands/webhook.js
```diff
@@ -0,0 +1,201 @@
...
+const decorateStatus = status =>
+	({
+		success: {
+			description: 'Biz-Ops powered Runbook validation: passed',
+		},
+		failure: {
+			description:
+				'Biz-Ops powered Runbook validation: runbook contains errors',
+		},
+	}[status]);
```
wheresrhys: bit too hipster? Would be easier to read if the map of descriptions was assigned to a module variable, and either a) the function grabs the right one b) does it even need a function if all it's doing is retrieving an entry from an object by name?

lambdas/ingester/serverless.yml
```diff
@@ -77,6 +80,23 @@ functions:
...
                   method: post
+    status:
+        description: Status page for GitHub checks
+        handler: src/statusPageHandler.handler
+        timeout: 30
+        events:
+            - http:
+                  path: /status/{owner}/{repo}/{hash}
+                  method: get
+    webhook:
```
wheresrhys: githubWebhook?

lambdas/ingester/src/templates/status-page.jsx
```diff
@@ -0,0 +1,61 @@
...
+					<RunbookEntry
+						placeholder={placeholder}
+						content={content}
+						readOnly={readOnly}
+					/>
+				</div>
+			</div>
+			<aside>
+				<p>
+					Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed
```
wheresrhys: I'm guessing you don't need an aside

lambdas/ingester/src/statusPageHandler.js
```diff
@@ -0,0 +1,45 @@
...
+				status: 200,
+				message: storedResult.message,
+				readOnly: true,
+				hash,
+				...storedResult.details,
+			},
+			event,
+		);
+	} catch (error) {
+		logger.info({
```
wheresrhys:
```suggestion
		logger.error({
```

lambdas/ingester/src/statusPageHandler.js
```diff
@@ -0,0 +1,45 @@
...
+			event: 'STATUS_PAGE_FAILED_FETCH_STORED_RESULT',
+			error: { ...error },
```
wheresrhys: why destructure?

lambdas/ingester/src/githubWebhookHandler.js
```diff
@@ -0,0 +1,6 @@
+const { serverless } = require('@probot/serverless-lambda');
```
wheresrhys: do we lose anything using this that using the lambda wrapper we use elsewhere gives us? Should we abstract any useful things out into a separate 'middleware' that we can use with probot or our lambda wrapper?

lambdas/ingester/src/commands/webhook.js
```diff
@@ -0,0 +1,201 @@
...
+	};
+
+	await Promise.all([
+		postStatus(ingestResult.status, statusDetails),
+		storeIngestResult(),
+	]);
+};
+
+exports.command = command;
+
```
wheresrhys: general comment on log sin this file:
- should use error instead of info in a few places
- lots of repetition which (because prettier spreading lines out) makes the file harder to read. Worth writing a logger wrapper that takes care of `repository, commitSha, runbookSha`?

lambdas/ingester/src/githubWebhookHandler.js
wheresrhys: NB - the answer may well be "No - we don't lose anything important"

lambdas/ingester/src/commands/webhook.js
doramatadora: I'll use an object and just-in-time resolution in the command

lambdas/ingester/src/templates/status-page.jsx
doramatadora: This template is WIP

lambdas/ingester/src/statusPageHandler.js
doramatadora: to avoid logging the methods on the error prototype

lambdas/ingester/src/commands/webhook.js
doramatadora: Can I address your second point as part of tech debt? I have an idea for this but it's going to take me a while to write

lambdas/ingester/src/statusPageHandler.js
wheresrhys: not sure if that's a problem. JSON.stringify (presumably used internally in the logger) ignores prototypes anyway... I think

doramatadora: but locally there's a lot of noise

PR #60, opened 2019-06-27
stuff for creating dynamo table

wheresrhys:
# Why
We need a persistence layer for runbook.md PR results

# What
cloudformation and make task for creating table - should not need to use again, but can be modified if we ever need to create other tables


Makefile
```diff
@@ -32,3 +32,13 @@ ifneq ($(CI),)
 else
 	jest --watchAll
 endif
+
+create-database:
+	aws cloudformation create-stack \
+	--region eu-west-1 \
+	--stack-name biz-ops-runbooks-md-data \
+	--template-body file://$$(pwd)/cloudformation/dynamodb.yaml \
+	--tags Key=description,Value="Surveys how much time we spend yak-shaving and dealing with bugs" \
```
doramatadora: Lies. I think this is a remnant from `yak-fly`

Limess:
Worth noting that CF doesn't persist tags properly to DynamoDB from my experience, you need to pass them as parameters instead and then set them explicitly, e.g. https://github.com/Financial-Times/healthcheck-metadata-store/blob/master/deployments/dynamodb.yml.

cloudformation/dynamodb.yaml
```diff
@@ -0,0 +1,23 @@
...
+        -
+          AttributeName: "Repository"
+          KeyType: "HASH"
+        -
+          AttributeName: "CommitHash"
+          KeyType: "RANGE"
+      ProvisionedThroughput:
+        ReadCapacityUnits: "1"
+        WriteCapacityUnits: "1"
+      TableName: "biz-ops-runbook-md.results"
```
Limess: I'd also add

```yaml
      SSESpecification:
        SSEEnabled: true
```

to enable encryption with the default amazon managed KMS keys.

Limess: That said the docs for that parameter aren't half confusing... it might be that it's still encrypted at rest if false, but with an entirely amazon managed key, rather than one that exists in your account? 😕 

> Indicates whether server-side encryption is done using an AWS managed CMK or an AWS owned CMK. If enabled (true), server-side encryption type is set to KMS and an AWS managed CMK is used (AWS KMS charges apply). If disabled (false) or not specified, server-side encryption is set to AWS owned CMK. 

https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-dynamodb-table-ssespecification.html

wheresrhys: Is too late now I guess. Can one add encryption to a table retrospectively?

No sensitive data anyway.

Limess: I suspect not...? It'd be in the UI if you can

doramatadora: There **IS** plenty of sensitive data @wheresrhys, because we store the validation results which can potentially include PII too 🤦‍♀ New encrypted table it is

PR #65, opened 2019-06-28
add a consistent error message for empty sections

chee:
Before this the error message for an empty section depended on the type of the section. so:

```
## contains personal data

## dependents

## more information
```

would print

```
line 1: i couldnt resolve  to a boolean
line 3: expected a list but didn't get any bulleted items
```

and now it would print:

```
line 1: property "containsPersonalData" has no value
line 3: property "dependents" has no value
```


GeoffThorpeFT:
Does this change mean than any empty section results in an error?  i.e. if a template is used which contains all the sections then the user will have to ensure they remove the unpopulated sections before submitting.

chee:
> Does this change mean than any empty section results in an error? i.e. if a template is used which contains all the sections then the user will have to ensure they remove the unpopulated sections before submitting.

Other than for subdocuments, any empty section always resulted in an error before. The error it resulted in was very confusing because it would be "`` is not a value for Enum` for enums, and stuff about bulleted lists for lists. 

This makes it consistent that every one of those is an error, and is the same error.

PR #70, opened 2019-07-04
Supply a list of properties to reject from runbook.md

GeoffThorpeFT:
## What
Error on the use of a selection of system properties from the runbook.md content.

## Why
Not all the system properties are best populated by the owner of the system. Some, such as dependents, are best populated by the owner of the item that uses the system.

## How
Created a property scope configuration.
Derived a simple list of properties to reject. If the property is found in the parsed result then an error message is added to the error list and the property is removed from the result.
Derived a simple list of properties to demand. If the property is not found in the parsed result then an error message is added to the error list.


GeoffThorpeFT:
image

lambdas/ingester/src/schema/property-scope.js
```diff
@@ -0,0 +1,58 @@
...
+};
+
+const matchingProperties = desiredScope =>
+	Object.entries(propertyScope)
+		.filter(([, scope]) => scope === desiredScope)
+		.map(([property]) => property);
+
+module.exports = {
+	propertyScope,
+	owned: matchingProperties('Owned'),
```
wheresrhys: Not sure if the discussion was captured in the trello ticket, but I think it's best to back away from the concept of Owned as there were too many properties that were borderline on this, which means we would get in a situation where the locking of fields by runbook.md starts to look quite arbitrary, and arbitrary changing of permissions is bound to get frustrating for users.

lambdas/ingester/src/schema/property-scope.js
```diff
@@ -0,0 +1,58 @@
...
+	recursiveDependencies: 'Excluded',
+	recursiveDependentProducts: 'Excluded',
+	recursiveDependents: 'Excluded',
+	releaseDetails: 'Owned',
+	releaseProcessType: 'Owned',
+	replacedBy: 'Excluded',
+	repositories: 'Excluded',
+	rollbackProcessType: 'Owned',
+	secondLineTroubleshooting: 'Owned',
+	serviceTier: 'Essential',
```
wheresrhys: Similar to my other comments on Owned, I feel increasingly uncomfortable with defining some things as Essential. The cutoff is always gonna be somewhat arbitrary e.g. why is serviceTier Essential, but lifecycleStage Owned?

wheresrhys:
Also, I think this validation should/could live in parser - a list of field names which make the runbook.md invalid is very similar in scope to the parser's validation against the schema. This list of invalid fields could be looked at as some extra schema metadata (maybe some day coudl move _into_ the schema)

GeoffThorpeFT:
I've added the solution to the parser as well (we can pick which solution is best):
image

chee:
> Also, I think this validation should/could live in parser - a list of field names which make the runbook.md invalid is very similar in scope to the parser's validation against the schema. This list of invalid fields could be looked at as some extra schema metadata (maybe some day coudl move _into_ the schema)
> 
> What d'you think @chee

yes

libraries/parser/lib/tree-mutators/validate-bizops-properties.js
```diff
@@ -1,8 +1,15 @@
 const visit = require('unist-util-visit-parents');
 const convertNodeToProblem = require('./convert-node-to-problem');
+const excluded = require('../../../../schema/excluded-properties');
 
 module.exports = function coerceBizopsPropertiesToType({ validateProperty }) {
 	function mutate(node) {
+		if (excluded.includes(node.key)) {
+			return convertNodeToProblem({
+				node,
+				message: `${node.key} is not permitted within runbook.md`,
```
wheresrhys:
```suggestion
				message: `${node.key} is not permitted within runbook.md as non-engineers or people in other teams may need to edit it`,
```

libraries/parser/lib/tree-mutators/validate-bizops-properties.js
```diff
@@ -1,8 +1,15 @@
 const visit = require('unist-util-visit-parents');
 const convertNodeToProblem = require('./convert-node-to-problem');
+const excluded = require('../../../../schema/excluded-properties');
```
wheresrhys: This file could move a lot closer to this file's location. Could even just be an array in this file.

What do you see as the advantages of keeping it at the root of the project?

libraries/parser/test/property-value.test.js
```diff
@@ -194,35 +213,52 @@ test('subdocuments have their headers reduced two levels', async () => {
...
-	expect(date.getMonth()).toBe(8);
-	expect(date.getDate()).toBe(17);
+	expect(errors).toHaveLength(1);
+	const [{ message }] = errors;
+	expect(message).toEqual(
+		'lastServiceReviewDate is not permitted within runbook.md',
+	);
 });
+//
+// test('date fields are coerced to iso strings', async () => {
```
wheresrhys: can you use test.skip instead of commenting out please? And leave a comment as to why they are skipped

PR #71, opened 2019-07-05
Increase releaseLog lambda timeout

OttlikG:
Trello ticket: [724](https://trello.com/c/NopGDyn8/724-increase-the-timeout-value-for-releaselog-lambda)

From 30 to 50 to potentially resolve the issue when
bizops api returned with the following

"message":"Waiting for Biz Ops Update response"


serverless.yml
```diff
@@ -80,7 +80,7 @@ functions:
     releaseLog:
         description: Apply runbook.md changes on prod release
         handler: lambdas/ingester/src/releaseLogHandler.handler
-        timeout: 30
+        timeout: 50
```
wheresrhys: m, definitely a bug somewhere as should not need longer than about a second. didn't realise we were already up to such high timeout values. I doubt this increase will help because something is probably just not terminating, so te effect of this will be we pay for 50s not 30s while it hangs.

Not sure what to suggest to investigate the bug though

OttlikG: I will close this for now, just wanted to raise awareness if this happenes in the furute.

PR #75, opened 2019-07-10
upgrading to node 10

jenniferbacon01:
## What
Upgrading to node 10

## Why
Keep up to date

## How
package.json, serverless.yml


jenniferbacon01:
> Cool, I was going to say CircleCI config needs doing as well but that was already using 10 🤔

Thought the circle CI config would just come from the serverless.yml?

PR #77, opened 2019-07-11
fix offline kinesis initialisation and bump deps

doramatadora:
# What
Check if local message stream is running / exists each time we `make run`, using make commands rather than variables

# Why
Fix docker errors when the container already exists


wheresrhys:
cool - I was concerned about the magic script Charlie had added - glad there's a clearer fix for it :+1:

PR #81, opened 2019-07-16
add serverless-offline-sts and region param to aws commands

jenniferbacon01:
## What
Adding region param to aws command and serverless-offline-sts plugin

## Why
So you don't need to have global aws credentials when running the app locally

## How
serverless.yml and Makefile


serverless.yml
```diff
@@ -125,6 +125,7 @@ plugins:
     - serverless-offline-kinesis
     - serverless-offline
     - serverless-prune-plugin
+    - serverless-offline-sts
```
GeoffThorpeFT: Not sure I understand when this is being added as its not used in any tests.

jenniferbacon01: @wheresrhys said that we should add it for potential future use

PR #84, opened 2019-07-17
add runbook score meter

doramatadora:
# What
Adds a visual indicator of the runbook operability score on the status page.
Screenshot 2019-07-17 at 13 02 52

# Why
Instant feedback on improvement / diminishment in runbook quality.


GeoffThorpeFT:
Nice, should we not round the percentage value to avoid showing the whole fractional element.

PR #88, opened 2019-07-22
strip html comments from runbooks

doramatadora:
# What

Strip HTML comments from raw runbook data before parsing.

# Why

HTML comments are harmless but they trip up the parser, which expects a heading on the first line, so the ingester errors for [some runbooks](https://github.com/Financial-Times/prometheus-health-check-exporter/blob/master/RUNBOOK.md).


wheresrhys:
belated :+1:

PR #91, opened 2019-07-24
Adding to readme

melanierogan:
# Why
I hit some snags when getting things running locally. Thankfully Dora was a huge help. I'm adding a few things here that should help make the process more smooth for the next person.  

# What
I added a line about ensure AWS CLI is installed, and a troubleshooting point if you get a `ENOTFOUND` error.


Limess:
Prettier runs on markdown files as part of `make verify` and the pre-commit hook, running `make verify` again locally should fix the build error.

melanierogan:
Ah brilliant, thanks!

PR #95, opened 2019-07-29
Log github API call failures

Limess:
Currently Github API failures are logged as destructuring errors, as the response is not in the expected array/object format. Anything is treated as a success.

This change propagates the child loggers to the github API caller, which logs an error and throws on any non ok status code.

Example error - which I believe implies that the returned value is not an array so is not iterable:

```
"error":{"type":"TypeError","message":"(intermediate value) is not iterable","stack":"TypeError: (intermediate value) is not iterable\n    at path (/var/task/lambdas/ingester/src/webpack:/lambdas/ingester/src/releaseLogHandler.js:110:9)\n    at process._tickCallback (internal/process/next_tick.js:68:7)"}
```


lambdas/ingester/src/releaseLogHandler.js
```diff
@@ -66,69 +66,97 @@ const parseRecord = childLogger => record => {
 	};
 };
 
-const githubAPI = (path, { previewMode, ...options } = {}) => {
+const createGithubAPI = (log = logger) => async (
```
wheresrhys: createGithubAPI is a bit of a weird name. Maybe createGitApiClient?

lambdas/ingester/src/releaseLogHandler.js
```diff
@@ -66,69 +66,97 @@ const parseRecord = childLogger => record => {
...
-};
+	if (!response.ok) {
+		const contentType = response.headers.get('content-type');
+		const body = await (contentType &&
+		contentType.includes('application/json')
+			? response.json()
+			: response.text());
+
+		log.error({
+			event: 'GITHUB_API_FAILURE',
```
wheresrhys: also log the url and other params sent to the api?

Limess: Good idea, added logging of the URL and all options aside from the Authorization header.

lambdas/ingester/src/releaseLogHandler.js
Limess: Added `Client` as a suffix.

PR #97, opened 2019-07-29
Support valid github repository names

Limess:
The current release log processing code doesn't support . or - characters in repository names. This results in incorrect parsing, e.g. runbook.md -> runbook.

This character list is taken from the one Docker use, so should be good enough for our use case (https://github.com/moby/moby/issues/679#issuecomment-18307522).


Limess:
Screenshot 2019-07-30 at 11 38 15

PR #98, opened 2019-07-30
Add an integration test for one of the happy-paths for release-log

Limess:
There weren't any tests for this handler.

Added tests using nock to stub all API calls, and calling the handler with a mock kinesis record.

This is relatively brittle due to the nature of the calls, and also involves stubbing the schema using a 'private'ish API as the library doesn't expose the ability to use stub rawData - this should be a thing.

I've used the prometheus-pingdom-exporter runbook as a fixture for now as it's relatively flushed out but this should probably be bespoke in future.

Additionally moved the babel config to be shared between webpack server/jest as jest blew up when getting near JSX.

~N.B this is branched from https://github.com/Financial-Times/runbook.md/pull/97 as it tests currently invalid repo names.~


babel.config.js
```diff
@@ -0,0 +1,27 @@
+'use strict';
+
+// server/test config
+module.exports = {
```
wheresrhys: Why did we get away without this before/why is it necessary now?

lambdas/ingester/src/lib/external-apis.js
```diff
@@ -17,7 +20,10 @@ const callExternalApi = async ({
 		body: JSON.stringify(payload),
 		headers,
 	};
-	const fetchResponse = await nodeFetch(url, options);
+	const fetchResponse = await nodeFetch(url, {
+		...options,
+		agent: keepAliveAgent,
```
wheresrhys: An optimisation, or does this fix something in particular? Are there any risks to using keep alive in a lambda?

lambdas/ingester/src/releaseLogHandler.js
```diff
@@ -217,7 +221,7 @@ const fetchRunbookMds = (parsedRecords, childLogger) =>
 				const [gitRepositoryName] =
 					gitRefUrl
 						.replace('https://github.com/', '')
-						.match(/[\w-]+\/[\w-]+/) || [];
+						.match(/[a-z0-9_.-]+\/[a-z0-9_.-]+/i) || [];
```
wheresrhys: Last time I checked, github had no published restrictions on characters in repo names. Arguably this regex should use `[^/]*`, but (unless you can find a reference for repo naming limitations), shouldn't be made struicter.

Or is this a git limitation, rather than github?

lambdas/ingester/src/releaseLogHandler.js
```diff
@@ -264,14 +268,22 @@ const fetchRunbookMds = (parsedRecords, childLogger) =>
 					},
 					'Retrieving runbook.md from GithubApi has failed',
 				);
+				throw Object.assign(
```
wheresrhys: this new error, I take it, is the meat of this PR?

lambdas/ingester/src/releaseLogHandler.js
```diff
@@ -304,7 +316,19 @@ const handler = async (event, context) => {
...
 
-	await processRunbookMd(parsedRecords, childLogger);
+	if (parsedRecords.length === 0) {
+		childLogger(
+			{
+				event: 'SKIPPING_INGEST',
+			},
+			'Nothing to ingest, skipping event',
+		);
+		return json(200, {
```
wheresrhys: along with this

lambdas/ingester/test/fixtures/runbook-fixture.md
```diff
@@ -0,0 +1,170 @@
+<!--
+    Written in the format prescribed by https://github.com/Financial-Times/runbook.md.
```
wheresrhys: this is actually not a bad thing to encourage people to do in general. We could add it to our example content

lambdas/ingester/test/releaseLogHandler.test.js
```diff
@@ -0,0 +1,261 @@
+const nock = require('nock');
```
wheresrhys: quibbling the naming of this test file - looks like there's a lot of github app type stuff in here

secret-squirrel.js
```diff
@@ -1,16 +1,16 @@
...
+		allow: [],
 		allowOverrides: []
 	},
 	strings: {
 		deny: [],
 		denyOverrides: [
-			'b62f4580-81f6-11e9-8f9b-7d694d159e85',
 			'b62f4580-81f6-11e9-8f9b-7d694d159e85', // README.md:9
-			'baxterthehacker@users.noreply.github.com'
+			'baxterthehacker@users\\.noreply\\.github\\.com', // githubPushEvent.json:20|25|42|47|60|130, lambdas/ingester/test/fixtures/githubPushEvent.json:20|25|42|47|60|130
```
wheresrhys: what's this guy's backstory? What's his motivation?

babel.config.js
Limess: Jest was never loading the client side files with JSX in them. As soon as it hits a code path that tries to load them (this handler for example) it blows up as it can't read the syntax.

Adding `babel-jest` and a config so it can read the files resolves this.

lambdas/ingester/src/lib/external-apis.js
Limess: Optimisation, probably premature.

No risks that I know of, if the socket is closed it'll just open a new one.

lambdas/ingester/src/releaseLogHandler.js
Limess: See https://github.com/Financial-Times/runbook.md/pull/97, there's a link to a moby (Docker) PR where they use this pattern.

I believe it's that github convert any other characters to these characters in URLs, if you have evidence otherwise then yes, `[^\/]` makes sense (I was going to use that prior to googling it).

lambdas/ingester/src/releaseLogHandler.js
Limess: One of the concrete changes yes. I wouldn't call it the meat though...

Just what I found while writing tests and trying to make them pass or give reasonable errors.

lambdas/ingester/src/lib/external-apis.js
wheresrhys: My worry is this might mean we get lots of weird, confusing-when-debugging socket closure error logs (we did in runbooks when doing some not dissimilar out of band requesting)

lambdas/ingester/src/releaseLogHandler.js
Limess: This doesn't actually do anything as nothing is inspecting the lambda output, but it does mean it's testable.

lambdas/ingester/test/fixtures/runbook-fixture.md
Limess: It's relatively common to have a header like this for machine generated files (puppet/ansible etc), I think it's a good idea to add.

lambdas/ingester/test/releaseLogHandler.test.js
Limess: It's matching the name of the handler file it's testing. There are a lot of naming issues in this repo (things not named as they are doing, inconsistent naming conventions in the same folder).

lambdas/ingester/src/lib/external-apis.js
Limess: I haven't seen this anywhere else - wasn't that down to making head requests or not reading the body on gets?

lambdas/ingester/src/releaseLogHandler.js
wheresrhys: yep they appear to. Frustrating it's not documented anywhere

lambdas/ingester/test/releaseLogHandler.test.js
wheresrhys: Ah in that case I suppose I'm surprised that the change log handler needs to interact with sos. or with pull requests. mind you, I haven't worked with the code much

lambdas/ingester/src/lib/external-apis.js
wheresrhys: yep, but the underlying problem was that sockets were not being closed, the lambda execution paused, and then the socket timed out. Same thing could happen here, no?

lambdas/ingester/test/releaseLogHandler.test.js
Limess: I was very surprised by how much this thing does in a single flow, I found it very hard to follow - one of the reasons I decided to write an end-to-end integration test for the handler.

There are actually more code paths which it can hit with github if it's not a PR which this test does not cover.

lambdas/ingester/src/lib/external-apis.js
Limess: I don't believe so as nothing happens in the background, it's all in the runtime of a single lambda.

If the lambda freezes and restarts after a set of records is processed I'm fairly sure that it'll get a new socket?

Limess:
Why a tentative 👍? Only related to adding keepAlive?

PR #102, opened 2019-07-31
fix status page view for runbooks with parse errors

doramatadora:
# What
* Do not display score tile if we couldn't calculate a runbook score (case: runbook contains parse errors)
* Fix missing error styling

# Why
Before
Screenshot 2019-07-31 at 12 26 56

After
Screenshot 2019-07-31 at 12 27 06


lambdas/ingester/src/templates/components/validation-result.jsx
```diff
@@ -34,11 +34,13 @@ exports.ValidationResult = ({
 				</p>
 			</aside>
 		)}
-		<div className="o-grid-row fullwidth with-margin-bottom">
-			<div data-o-grid-colspan="12">
-				<ScoreTile value={weightedScore} />
+		{!parseErrors.length && (
```
wheresrhys: If there are parseErrors will this output the number?

doramatadora: It's output below, in the error message - this is just to hide the score tile, which is NaN (not calculated)

PR #109, opened 2019-08-08
first pass at auto export tool

wheresrhys:
# Why
I was gonna write a runbook.md for biz-ops-runbooks, and didn't want to start from scratch.

# What
Bit hacky but
- gets content from biz ops
- avoids blacklisted fields
- dumps it all out in - hopefully - a valid runbook.md


lambdas/ingester/src/exportHandler.js
```diff
@@ -0,0 +1,71 @@
...
+		.replace(/([a-z])([A-Z])/g, '$1 $2')
+		.replace(/^[a-z]/, $0 => $0.toUpperCase());
+
+const handler = async event => {
+	const { excludedProperties } = runbookMd(schema);
+	const { systemCode } = event.queryStringParameters;
+
+	const data = await readSystem(systemCode);
+	excludedProperties.forEach(prop => {
+		if (prop in data) {
```
doramatadora: I think the `if` statement is redundant here

libraries/parser/lib/tree-mutators/validate-bizops-properties.js
```diff
@@ -18,6 +18,9 @@ const excludedProperties = [
 	'sosTrafficLight',
 	'stakeholders',
 	'updatesData',
+	'dataOwner',
```
doramatadora: I'm just curious why we are excluding the GDPR related fields

wheresrhys: because non tech compliance people tend to have at least some say in what these values are. Locking the fields by a doc in github makes tech the gatekeepers of this info, and they shouldn't be

lambdas/ingester/src/exportHandler.js
wheresrhys: In earlier versions of javascript it used to issue a warning, I'm fairly sure, but does seem redundant now

PR #114, opened 2019-08-14
Export tool 3

wheresrhys:
A few minor tweaks to the presentation


lambdas/ingester/src/lib/generate-from-biz-ops.js
```diff
@@ -39,10 +42,10 @@ exports.generate = async systemCode => {
     Written in the format prescribed by https://github.com/Financial-Times/runbook.md.
     Any future edits should abide by this format.
 -->
+
```
wheresrhys:
```suggestion
```

PR #128, opened 2019-08-19
Generate github issue

melanierogan:
# Why
With runbook.md running on each PR, if there's a failure with parsing or validation, folks don't immediately know there is a problem. So this PR goes about posting a github issue tagging the user so folks know when there is an issue with their runbook. 

# What
Added `postRunbookIssue` function that opens an issue in the relevant repo at point of failure, `releaseLogHandlerFailure` test that covers this.


lambdas/ingester/src/commands/ingest.js
```diff
@@ -91,7 +91,7 @@ const ingest = async (username, payload) => {
 	);
 
 	if (status !== 200) {
-		return decorateError({
+		throw decorateError({
```
doramatadora: 👍

lambdas/ingester/src/releaseLogHandler.js
```diff
@@ -128,6 +128,19 @@ class RunbookSource {
 		this.githubAPI = githubAPI;
 	}
 
+	async postRunbookIssue(gitRepositoryName, error, user) {
```
doramatadora: The second and last args could have less ambiguous names, like `errorCause` and `userGithubName` (you could destructure the user's github name rather than pass in the entire object)

lambdas/ingester/src/releaseLogHandler.js
```diff
@@ -128,6 +128,19 @@ class RunbookSource {
...
+		const path = `/repos/${gitRepositoryName}/issues`;
+		const requestOptions = {
+			method: 'POST',
+			body: JSON.stringify({
+				title: 'Runbook Error',
+				body: `@${user.githubName} There was an error with your runbook.md: ${error}. You can find further details on what went wrong on splunk: index=aws_cloudwatch source=*runbook-md* {insert system code here}`,
```
doramatadora: We know the system code from the release log payload. It's used in [other places in the code](https://github.com/Financial-Times/runbook.md/pull/128/files#diff-c49cdc4ba78f9693d4741db78f374fb4R199)

lambdas/ingester/src/releaseLogHandler.js
```diff
@@ -336,22 +362,43 @@ const getRecordIDs = records => records.map(({ eventID }) => eventID);
...
 		const ingestedRunbooks = await ingestRunbookMDs(
 			runbookMDs,
 			childLogger,
 		);
+		// console.log(ingestedRunbooks);
+
+		// commenting out the below because we never get to this point
+		// extra code that was confusing
 
 		const errors = ingestedRunbooks.filter(result => result.status >= 400);
```
doramatadora: should this be `result.status !== 200` to mirror the `ingest` function's output?

lambdas/ingester/src/releaseLogHandler.js
```diff
@@ -128,6 +128,19 @@ class RunbookSource {
 		this.githubAPI = githubAPI;
```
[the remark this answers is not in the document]
jenniferbacon01: Also feel it might read better if this class was in another file?

lambdas/ingester/src/releaseLogHandler.js
```diff
@@ -179,25 +192,28 @@ class RunbookSource {
...
+			.map(async ({ user, systemCode, repository, content, eventID }) => {
 				const fetchRunbookLogger = childLogger.child({ eventID });
 
 				try {
-					if (!user) {
-						throw new Error(
-							'Missing FT username when ingesting runbook',
-						);
-					}
 					fetchRunbookLogger.info({
```
jenniferbacon01: Can `fetchRunbookLogger.info(......)` still throw an error or was the try catch block just for the error above?

lambdas/ingester/test/releaseLogHandlerFailure.test.js
```diff
@@ -0,0 +1,277 @@
...
+		baseUrl: global.process.env.SCHEMA_BASE_URL,
+		updateMode: 'dev',
+	});
+	return schema;
+});
+
+describe('Release log handler', () => {
+	beforeAll(() => {
+		nock.disableNetConnect();
+		nock.enableNetConnect('127.0.0.1');
```
jenniferbacon01: Could maybe add a comment to say what this does?

lambdas/ingester/test/releaseLogHandlerFailure.test.js
```diff
@@ -0,0 +1,277 @@
...
+	afterEach(() => {
+		nock.cleanAll();
+	});
+
+	afterAll(() => {
+		nock.enableNetConnect();
+		jest.restoreAllMocks();
+	});
+
+	const addGithubPullRequestInterceptor = ({
```
jenniferbacon01: I would maybe move all these functions into a separate file so the test looks a bit cleaner - it's up to you though! Depends on your own style :-)

lambdas/ingester/test/releaseLogHandlerFailure.test.js
```diff
@@ -0,0 +1,277 @@
...
+		repositoryName,
+		pullRequestNumber,
+		modifiedRunbookSha,
+	}) =>
+		nock('https://api.github.com')
```
jenniferbacon01: You can take the below calls out into a const for DRY reasons (which I always think should be DNRY for Do Not Repeat Yourself but its less catchy I guess)
```
nock('https://api.github.com')
.defaultReplyHeaders({
'Content-Type': 'application/json',
})
.replyContentLength()
.replyDate()

File: lambdas/ingester/src/releaseLogHandler.js [hunk 10]
```diff
@@ -336,22 +362,43 @@ const getRecordIDs = records => records.map(({ eventID }) => eventID);
 
 const processRunbookMd = async (parsedRecords, childLogger) => {
 	const eventIDs = getRecordIDs(parsedRecords);
+	const runbookSource = new RunbookSource({
+		githubAPI: createGithubAPIClient(),
+	});
 	try {
+		// first point of failure: if this errors, we go into the catch block
+		// this can only error once – Promise.all bails as soon as one its promises rejects
```
wheresrhys: So in a scenario where 2 PRs on 2 repos are merged at more or less the same and both change events are in the same kinesis batch, then if one of them fails the other will fail too? is that right? That's not a great way to handle that failure as we have problems caused by one user impacting the success of a completely unrelated user

It's not a problem introduced in this PR, but a while ago I was thinking we could refactor so that there are 2 things

- a 'pipeline' which takes a single record and goes as far as it can - ideally all the way - towards successfully ingesting into Biz Ops
- A 'runner' function, which is essentially:

```js
const runner = kinesisRecords =>
	Promise.all(
		records.map(record => 
			doTheEndToEndIngestionForOneRecord(record)
				.catch(notifyTheUserOfFailure)
		)
	)
```

Might be a good time to try that as it'd make error handling more streamlined as well as avoiding failures being leaky

lambdas/ingester/test/fixtures/runbook-fixture-bad.md
```diff
@@ -0,0 +1,168 @@
+<!--
+    Written in the format prescribed by https://github.com/Financial-Times/runbook.md.
```
wheresrhys: For ease of maintenance I would go for a pared down runbook.md. Aside from the H1 at the top, pretty much none of it is required. A good question to ask when writing tests & test data is 'What's the minimimum set of data I need in order to recreate the effect I'm after?'

lambdas/ingester/test/releaseLogHandlerFailure.test.js
```diff
@@ -0,0 +1,277 @@
...
+						name: 'Reliability Engineering',
+						email: 'reliability.engineering@ft.com',
+						slack: 'reliability-eng',
+						phone: 'n/a',
+						supportRota: 'rota',
+						contactPref: 'Slack #reliability-eng',
+						isActive: true,
+						productOwners: [
+							{
+								code: 'sarah.wells',
```
wheresrhys: ah, I remember us discussing in the last few weeks that it'd be good to scrub real people out of this test data

lambdas/ingester/src/releaseLogHandler.js
```diff
@@ -77,6 +70,24 @@ class RunbookSource {
...
+		githubName,
+		errorCause,
+		systemCode,
+	) {
+		const path = `/repos/${gitRepositoryName}/issues`;
+		const requestOptions = {
+			method: 'POST',
+			body: JSON.stringify({
+				title: 'Runbook Error',
+				body: `@${githubName}, there was an error with your runbook.md: ${errorCause}. You can find further details on what went wrong on splunk: index=aws_cloudwatch source=*runbook-md* ${systemCode}`,
```
doramatadora: I think the title and message can be a little more telling, e.g. `runbook.md automated runbook ingestion failure`
```
... there was an error synchronising runbook data with Biz-Ops: {errorCause}.
This automated operation was triggered by a recent release
[do we want to include any data about the release here?]
logged in [Change API](link).
You can find further details about what went wrong on [Splunk](link): ... 
Issue posted automatically by [runbook.md](link to docs)
```

lambdas/ingester/src/releaseLogHandler.js
```diff
@@ -195,6 +206,7 @@ const fetchRunbookMds = (parsedRecords, childLogger) =>
 			});
 
 			let repository = 'UNKNOWN';
+			// console.log('*********git url here**********', gitRefUrl);
```
doramatadora: Can this comment be removed?

lambdas/ingester/src/releaseLogHandler.js
melanierogan: Much better, I've adjusted to fit this now. I haven't put all data about the release and instead pointed to the splunk logs. I'm happy to adjust this if you feel it is more useful to have the release info in the issue itself.

PR #135, opened 2019-09-04
clearer info on irrlevant poroperties in generated r.md

wheresrhys:
# Why
Feedback from Euan and chris Dunham that the placeholder messages cause errors when they are uncompletable

# What
Clearer guidance on removing the heading and placeholder if irrelevant


wheresrhys:
Got a :+1: IRL from Euan

PR #138, opened 2019-09-05
GitHub issue logging

melanierogan:
# What
Adding logging for the `postRunbookIssue` function. Also, adding in the 'runbooks-in-repos' slack channel for folks who have questions. 

# Why
The `postRunbookIssue` function is silently failing and I need to figure out why so I can debug. The slack channel was added based on feedback on issue copy from the team.


lambdas/ingester/src/releaseLogHandler.js
```diff
@@ -76,20 +76,37 @@ class RunbookSource {
...
+				);
+				err.response = postIssue;
+				throw err;
+			}
+
+			return postIssue;
+		} catch (error) {
+			logger.info({
+				event: 'GITHUB_POST_FAILED',
+				error,
```
doramatadora: It would be good to log `gitRepositoryName` too

melanierogan: Good shout, 'll add that now

PR #141, opened 2019-09-06
taking out async and await in postRunbookIssue

melanierogan:
Sorting out the postRunbookIssue which doesn't need async, or logging now 

Edit: 
# Why
We failed to notice GitHub errors are handled and logged separately a few lines above

# What
Removing logging from `postRunbookIssue`, async/await and just returning without variable assignment.


lambdas/ingester/src/releaseLogHandler.js
```diff
@@ -70,44 +70,22 @@ class RunbookSource {
...
-		} catch (error) {
-			logger.info({
-				event: 'GITHUB_POST_FAILED',
-				error,
-				gitRepositoryName,
-			});
-		}
+			}),
+		};
+		const postIssue = this.githubAPI(path, requestOptions);
```
doramatadora: you can just return here and skip the variable assignment

PR #150, opened 2019-10-03
Retrieve system code from runbook

fenglish:
## Why?
If a repo contains system a and system b, and a change event is sent for system a, should both runbooks be updated?
[The trello ticket](https://trello.com/c/qxpEVk2X/71-clarify-relationship-between-change-api-events-and-runbook-updates-when-multiple-runbooks-exist-in-1-repo)

## What has changed?

- Retrieve systemCode from runbook.md

  **`systemCode` validate priority**
   1. System code from runbook.md contents
**(👆 this part is added by this PR to deal multiple runbooks)**
   2. From .github/runbooks.yml config, if it exists (this is shared with github app)
   3. From file name system-code_runbook.md (convention)
   4. the one in the Change API message

  
- Add an api call to check whether the systemCode exists in BizOps before update

-   [ ] Bug fix
-   [x] Feature
-   [ ] Breaking change
-   [ ] Review not required

## Reviewer checklist


lambdas/ingester/src/commands/ingest.js
```diff
@@ -26,13 +27,28 @@ const decorateError = props => {
...
+		let message;
+		let code;
+		if (e.status === 404) {
+			message =
+				'Parse & validation complete. Biz-Ops update skipped (systemCode not exist in BizOps).';
+			code = 'parse-ok-systemCode-not-exist';
+		} else {
+			message =
+				"Parse & validation complete. Biz-Ops update skipped (BizOps api error when check systemCode's existence).";
+			code = 'parse-ok-biz-ops-api-error';
```
doramatadora: What do you think about:
-  `Biz-Ops update skipped: system code not found.` & `parse-ok-system-code-not-found`
-  `Parse & validation complete. Biz-Ops update skipped. Error from Biz-Ops: {e.message}.` – expose the specifics of the error from Biz-Ops API

ysugimoto: I think exposing the error from Biz-Ops API would be better.

When we call the API with an invalid or not found system code, the Biz-Ops API responds with an error message like `System [system-code] does not exist` . it is easy to catch what system code we sent to Biz-Ops API.

lambdas/ingester/src/commands/ingest.js
```diff
@@ -3,6 +3,7 @@ const runbookMd = require('../lib/parser');
 const { validate, updateBizOps } = require('../lib/external-apis');
 const { updateSystemRepository } = require('../lib/biz-ops-client');
 const { transformCodesIntoNestedData } = require('../lib/code-validation');
+const { readSystem } = require('../lib/biz-ops-client');
```
doramatadora: Oooh just noticed, can this be rolled into the import on line 4, so we don't require the file twice?

fenglish: Oh I see! Thank you! I will do that

PR #152, opened 2019-10-07
Systemcode existence on release handler

ysugimoto:
## Why?

We should prioritize a system code detection on `relaseLogHandler` as following order:

1. System code from runbook.md contents (ref: #150)
2. From .github/runbooks.yml config, if it exists (this is shared with github app)
3. From file name system-code_runbook.md (convention)
4. the one in the Change API message

But this is a part of #150 implementation stuff.

## What has changed?

This PR implements system code detection on `releaseLogHandler` same as `githubWebhookHandler`.  

Before send a ingest, we'll use a system-code which is prioritized:

1. Fetch `.github/runbooks.yml` from respository (if exists) and find a system-code which corresponds to changed path. If found, use it
2. If `[system-code]_runbook.md` is found in PR/Commit changes, parse and use [system-code] but it'll use convention one, it means _use system code which is found first_
3. Otherwise, use system-code which come from Change-API message. 

To parse `.github/runbooks.yml`, I use `js-yaml`. If we have a better way to parse md, let me know (Does it prefer to use AST as same as runbook.md parsing?)

-   [ ] Bug fix
-   [x] Feature
-   [x] Breaking change
-   [ ] Review not required

## Reviewer checklist


lambdas/ingester/src/releaseLogHandler.js
```diff
@@ -87,12 +89,35 @@ class RunbookSource {
 		return this.githubAPI(path, requestOptions);
 	}
 
+	// Fetch configured runbook.yml under .github/runbook.yml
+	async getSystemCodeFromRepositoryConfig(gitRepositoryName) {
+		const path = `/repos/${gitRepositoryName}/contents/.github/runbooks.yml`;
```
ysugimoto: Should add exact commit hash of the current commit

lambdas/ingester/src/releaseLogHandler.js
```diff
@@ -87,12 +89,35 @@ class RunbookSource {
...
 
+	// Fetch configured runbook.yml under .github/runbook.yml
+	async getSystemCodeFromRepositoryConfig(commit, gitRepositoryName) {
+		const path = `/repos/${gitRepositoryName}/contents/.github/runbooks.yml?ref=${commit}`;
+		try {
+			const { content } = await this.githubAPI(path);
+			const obj = yaml.safeLoad(decodeBase64(content));
+			return (obj.runbook || {}).systemCodes || {};
+		} catch (e) {
+			this.logger.info(
```
fenglish: @ysugimoto I'm not sure we need this log because github handler's one doesn't do that.

wheresrhys:
What's the state of documentation of where system code to runbook.md mappings come from? There are now at least 5 potential sources as far as I can see:
- in the runbook
- name of runbook
- runbook.yaml config in the .github directory
- link between repository and system stored in biz-ops
- system code sent in change-api event

I don't know which of these are used and take precedence if there's a conflict

wheresrhys:
I'm not entirely sure how this solves the problem outlined in the ticket https://trello.com/c/qxpEVk2X/71-clarify-relationship-between-change-api-events-and-runbook-updates-when-multiple-runbooks-exist-in-1-repo. I'm not suggesting it _doesn't_ solve the problem, but nothing in the PR description explains what decisions have been made in response to  the questions in the ticket. The code looks technically good, but it's hard to review without knowing the details of what behaviour it is intended to implement

What's the state of documentation of where system code to runbook.md mappings come from? There are now at least 5 potential sources as far as I can see:
- in the runbook
- name of runbook
- runbook.yaml config in the .github directory
- link between repository and system stored in biz-ops
- system code sent in change-api event

There's a code comment which addresses which of these are used and take precedence if there's a conflict, but this should be documented for the end user, not just in a code comment

lambdas/ingester/src/releaseLogHandler.js
ysugimoto: @fenglish I agree with you because this logging is purpose for debugging. I'' remove this

fenglish:
@wheresrhys This PR doesn't include everything for the solution because this is a working branch between me and @ysugimoto towards a draft PR(#150). We will update the README and tidy up the #150 PR when all change were done. We should have wrote that this PR is working PR(?) for us. Sorry 😞

ysugimoto:
@wheresrhys As @fenglish says, This PR is a part of #150, but indeed, it's hard to review without decision background, sorry. I'll update the PR description and where is the better place to document it? `README.md` is good for us?

lambdas/ingester/src/releaseLogHandler.js
doramatadora: No, this should always be the mapping from the config in the primary branch. That's the behaviour of the github app (see this "won't fix" from the Probot team: https://github.com/probot/probot/issues/561)

wheresrhys:
Thank you both. I'm gonna remove the CODEOWNERS file as it tagged me as a reviewer unnecessarily.

Best place to document the system code source hierarchy would be a decision for Dora. We did for a while plan a documentation tidy up, but not sure how far that got

doramatadora:
We need a new section in the `README` about system codes relating to systems.

This PR means feature parity for release log & github app handlers in terms of where we look for system codes:

1. system code specified in runbook file (heading `Code`)
2. system code specified in `.github/runbooks.yml`
3. system code in path `**/my-system-code_runbook.md`
4. system code in change api event [Change API only]

@fenglish is also working on validating system codes before we update biz-ops, so we'll stop create new systems by mistake.

@wheresrhys we don't consider `link between repository and system stored in biz-ops` anywhere in the release log / github app toolchain – and never have in the past, because we agreed not to:
- it doesn't solve the problem of multiple runbooks in a repo
- if anything, that link in Biz-Ops should be updated automatically based on runbook / Change API data

lambdas/ingester/src/releaseLogHandler.js
doramatadora: Config is optional – I know of only very few systems with a manual / semi-automated release process that use this so agree with @fenglish on avoiding logging here

lambdas/ingester/src/releaseLogHandler.js
```diff
@@ -87,12 +89,35 @@ class RunbookSource {
...
+		}
+	}
+
 	getRunbookUrlFromModifiedFiles(files) {
 		const firstRunbookFound = files.find(({ filename }) =>
 			/runbook\.md$/i.test(filename),
 		);
-		const { contents_url: runbookUrl } = firstRunbookFound || {};
-		return runbookUrl;
+		const { contents_url: runbookContentUrl, filename } =
```
doramatadora: This is going to be a more involved refactor, I'm afraid. We want to find **all the runbooks** _in a repo_ (1 repo per change api event) `that have been affected by the commit`, not just the first one, and then match them up against system codes:

1. from the runbook contents themselves [individual]
2. from `.github/runbooks.yml` in the repo, **on the primary branch!**, so we need a cache of system codes we can search in [common]
3. from the file path [individual]
4. from the change-api event [common]

doramatadora: Think `Array.filter` instead of `Array.find`

doramatadora: 1. We should deal with each runbook found in the exactly same way we deal with a single now – that includes ingest and will require thinking carefully about how we handle errors (say we have 3 runbooks wrapped up in the same commit/pr/tree/whatever, and only 1 is valid, do we need to do anything extra in terms of error handling?)
2. The primary branch is what you get when you don't specify a branch when you make a request to the GitHub API, so just search for the file path (`.github/runbooks.yml`, always) rather than a specific commit or ref
3. Cache makes it sound so sofisticated, I just wanted to say, make sure we don't look up this file every time we look up a system code in a repo, if there's more than 1 runbooks in that repo. Can be as basic an implementation as an object outside of the runbook processing loop.

lambdas/ingester/src/releaseLogHandler.js
```diff
@@ -1,14 +1,21 @@
...
 const { parseKinesisRecord } = require('./lib/kinesis-util');
+const {
+	getMappedSystemCode,
+	parseSystemCode,
+	runbookRx,
+} = require('./lib/system-code');
 
 const githubApiUrl = `https://api.github.com`;
 const bizOpsApiKey = process.env.BIZ_OPS_API_KEY;
+const runbooksConfigurationCaches = {};
```
ysugimoto: Declare cache object, the key is repository name

lambdas/ingester/src/releaseLogHandler.js
```diff
@@ -87,12 +94,33 @@ class RunbookSource {
...
-			/runbook\.md$/i.test(filename),
-		);
-		const { contents_url: runbookUrl } = firstRunbookFound || {};
-		return runbookUrl;
+		return files
+			.filter(({ filename }) => runbookRx.test(filename))
+			.map(({ contents_url: runbookContentUrl, filename }) => ({
+				runbookContentUrl,
+				filename,
+			}));
```
ysugimoto: Returns array in order to deal with multiple runbooks if found

lambdas/ingester/src/releaseLogHandler.js
```diff
@@ -239,7 +298,18 @@ const fetchRunbooks = async (parsedRecords, childLogger) => {
...
+		.reduce((runbooks, next) => [...runbooks, ...(next || [])], [])
+		.reduce((runbooks, next) => {
+			const exists = runbooks.find(
+				runbook => runbook.systemCode === next.systemCode,
+			);
+			if (!exists) {
+				runbooks.push(next);
+			}
+			return runbooks;
+		}, []);
```
ysugimoto: In order to guard against duplicated system code contains, I added this function. This function drops duplicated system codes in fetched runbooks (if exists)

lambdas/ingester/src/releaseLogHandler.js
```diff
@@ -320,6 +390,8 @@ const processRunbookMd = async parsedRecords => {
 
 		if (!ingestedRunbooks.length) {
 			throw new Error('Bailing, could not ingest any runbooks.');
+		} else if (ingestedRunbooks.length !== runbookInstances.length) {
+			throw new Error('Bailing, either of runbook could not ingest.');
```
ysugimoto: Extra error handling, if either of ingest is failed, we should bail

lambdas/ingester/src/releaseLogHandler.js
fenglish: I wonder if just dropping duplicated system codes is safe or not. We should not update the runbook if there is a duplication in the PR or commit, because it may end up updating a wrong runbook? Although I don't  understand the situation completely... :see_no_evil:

lambdas/ingester/src/releaseLogHandler.js
fenglish: Don't we need to consider the situation when the config is changed? I know it won't be changed often.

ysugimoto: No we don't. This is because some changes come from Kinesis message, it's hard to find each `.github/runbooks.yml` changes for each messages. So we should get it from *primary branch* only once and use it on the whole process. I assume that the `.github/runbooks.yml` change should be treated as changing project configuration.

lambdas/ingester/src/releaseLogHandler.js
ysugimoto: In my understanding, the Kinesis message will come per PR/Commit so typically these won't conflict... I guess that the `ingest` will send per repository, not a commit/pr [here](https://github.com/Financial-Times/runbook.md/pull/152/files#diff-c49cdc4ba78f9693d4741db78f374fb4L245-L252) so it would be safe to keep uniqueness I thought.

On the other hand, you might be correct which we shouldn't keep uniqueness, then this function doesn't need. @doramatadora Could you help me with this?

lambdas/ingester/src/releaseLogHandler.js
doramatadora: Yep, that's how the github app works – the config is parsed from `.github/runbooks.yml` from the primary branch (usually `master`) at the time of the commit/PR.

lambdas/ingester/src/releaseLogHandler.js
doramatadora: I like your thinking here @ysugimoto, but I reckon it's not up to us to make these assumptions.

@fenglish here's an example of when this might happen: 
- `runbook.md` has `code: my-system`
- `other-system/runbook.md` has a mapping in config `my-system: other-system/runbook.md`
- say `my-system_runbook.md` is also in the repo, without any identifiable config

All these runbooks would have the same inferred system code of `my-system`. This is an exaggerated, but completely possible, scenario. In this case what the app should be doing is a) bailing for conflicting runbooks and not updating anything in biz-ops and b) alerting the user (by creating an issue on the repo) that their set-up is a mess.

Either way, it's not an edge case for this PR, nor is it high priority, so I'd remove the code and create a ticket on Trello instead.

lambdas/ingester/src/releaseLogHandler.js
doramatadora: Why? The current mechanism is creating an issue on the repo for failed ingests. It should be adapted so that it reports which system codes / runbooks were not ingested, but say we have 5 runbooks and 3/4 ingests succeeded so far, there's no reason to not attempt the 5th one. Perhaps I've misunderstood?

lambdas/ingester/src/releaseLogHandler.js
fenglish: I see 👍

lambdas/ingester/src/releaseLogHandler.js
ysugimoto: On this line, all runbooks seems to be ingested ( either failed or not ), so that is just checking all result of failed ingests. But as you say all failed ingests will be reported so I don't need to bail...

lambdas/ingester/src/releaseLogHandler.js
ysugimoto: I agree with that. a is safer but b is user friendly I think.

lambdas/ingester/test/releaseLogHandler.test.js
```diff
@@ -348,4 +393,232 @@ describe('Release log handler', () => {
...
 	});
+
+	it('should use configured system code at .github/runbooks.yml', async () => {
+		const givenSystemCode = 'biz-ops-runbook-md-from-change-api';
+		const givenRepositoryName = 'ft-repo.com.1';
+		const givenPullRequestNumber = '89';
+		const givenModifiedRunbookSha =
+			'bf8ddcc3b47ef8947fbfcbd84f0e231e4eade4cb';
+
+		const givenEvent = kinesisFixture.get({
```
doramatadora: This test has a lot of fixtures that could be extracted (in a function?) to avoid repetition

ysugimoto: That's good idea! I'll do in other PR

PR #157, opened 2019-10-11
Connect systems to repositories

ysugimoto:
## Why?

When a runbook.md is ingested from a repository we can infer that the `HAS_REPOSITORY` relationship should exist between the system and the repository in neo4j

However, this cannot be set using the main biz ops write because that uses `relationshipAction=replace`, whereas with repositories we will want `relationshipAction=merge` (because a system may have multiple repositories, not just the current one)

So to satisfy this requirement, ingestion should end with a second biz ops call, writing the relationship to the repository using the `merge` option

## What has changed?

I implement the feature of sending additional `PATCH` request after the main biz-ops update api call has succeeded.

I add new `updateSystemRelationships` function which intends to call biz-ops api with following endpoint and parametaters:

```
PATCH ${BIZ_OPS_API_BASE_URL}/v2/node/System/${systemCode}?relationshipAction=merge
{
  "repositories": ["github:${repositoryName}"]
}
```

If the `updateSystemRelationships` API fails, Is it OK to bail? or we should do this as optional request that respond 200 even if it fails

-   [ ] Bug fix
-   [x] Feature
-   [ ] Breaking change
-   [ ] Review not required

## Reviewer checklist


lambdas/ingester/src/commands/ingest.js
```diff
@@ -1,12 +1,14 @@
 const stripHtmlComments = require('strip-html-comments');
 const runbookMd = require('../lib/parser');
 const { validate, updateBizOps } = require('../lib/external-apis');
+const { updateSystemRelationships } = require('../lib/biz-ops-client');
```
wheresrhys: I think this and related parts of the code should be called `updateSystemRepository` - the main runbook.md ingest already updates lots of other relationships (e.g. dependencies), so it's clearer that this new functionality explicitly refers to repositories

lambdas/ingester/src/commands/ingest.js
```diff
@@ -1,12 +1,14 @@
...
 const { transformCodesIntoNestedData } = require('../lib/code-validation');
 
 const transformIngestedDetails = (
 	parseResult,
 	validationResult,
 	writeResult,
+	updateRelResult,
```
wheresrhys: similar to my comment above, this variable should reference repositories in its name

ysugimoto:
@wheresrhys Thanks! I'll merge this

wheresrhys:
You can create a tagged release too

ysugimoto:
OK - next tag will be `v51`

PR #158, opened 2019-10-11
Add first comment line numbers to validation report's line number

fenglish:
## Why?

If a runbook.md has a html comment at the top of it of N lines, then all the line numbers referred to in the runbook validation report are out by N.
[The trello ticket](https://trello.com/c/xT7KoHtM/97-line-numbers-in-runbook-report-dont-count-comments)

This PR only fixes the issue which the first comment lines are not counted in the validation report lines. There are still uncounted comments lines.

## What has changed?

- Add countFirstCommentLines function
- Pass firstCommentLines property to parse result forms

-   [x] Bug fix
-   [ ] Feature
-   [ ] Breaking change
-   [ ] Review not required

## Reviewer checklist


lambdas/ingester/src/commands/ingest.js
```diff
@@ -4,6 +4,8 @@ const { validate, updateBizOps } = require('../lib/external-apis');
 const { updateSystemRepository } = require('../lib/biz-ops-client');
 const { transformCodesIntoNestedData } = require('../lib/code-validation');
 
+let firstCommentLines;
```
wheresrhys: firstCommentLineCount would make it clearer that this is a n integer, rather than a copy of the contents of those lines

wheresrhys: Also, in general having module level variables like this (i.e. a variable every function in the module has access too) can be dangerous. it's not so much of a problem in a lambda as only 1 invocation at once can run at once in a lambda container (unlike on heroku/in an express app), but it's still good practice to, wherever possible, have variables scoped within a function. If another function needs access to the same variable, then pass it in to the function as a parameter/option.

fenglish: Thank you for your comment! I will modify them.

doramatadora:
I'm not sure this is the right approach – we don't just strip the comment at the top of the runbook, we in fact strip all HTML comments from the file. I propose using a 
with [`RegExp.exec`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/RegExp/exec) or [`String.match`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/match) with the `/g` flag – the line numbers apply to the fragments in between comments so we'd have to keep a running counter outside of the loop, and increment this with the length of the fragments (comments) found.

doramatadora:
@wheresrhys – I still don't think this is the right approach

fenglish:
This PR will solve all comments line number issue not only first one https://github.com/Financial-Times/runbook.md/pull/160

PR #167, opened 2019-10-30
No biz ops get

wheresrhys:
## Why?
This app is the last consumer of Biz Ops GET api, which will soon be switched off

## What has changed?
- In the runbook builder replaced GET with graphql, and also cleaned up things, hopefully squishing the fact that some deprecated properties and weird stuff were showing up in generated runbook.md's
- In ingest, used a head request to check if a system exists. This may become redundant as a) only applies to change log handler (I think)... adn that will very soon send systemIsInBizOps data, so no call to the api needed to find out

Not really

-   [x] Tech hygiene (dependency updating & other tech debt)
-   [ ] Bug fix
-   [ ] Feature
-   [ ] Breaking change
-   [ ] Review not required

## Reviewer checklist


lambdas/ingester/src/lib/biz-ops-client.js
```diff
@@ -30,28 +38,27 @@ const logAndThrowError = async (response, props) => {
...
+		body: JSON.stringify({ query, variables }),
+	}).then(async response => {
+		if (!response.ok) {
+			await logAndThrowError(response, {
+				method: 'graphql',
+			});
+		}
+		const json = await response.json();
+		if (json.errors) {
+			logAndThrowError(json);
```
doramatadora: I think you're missing an `await` here – this function will still return a promise.

wheresrhys: I ended up refactoring the error handling to make it a bit less 'one error handler function to rule them all' - there are 3 quite different cases
- turn an error response returning JSON into an error
- turn a 200 graphql response that has query errors in the body into an error
- turn a HEAD error response, with errors in the headers, into an error

It's now split up into separate functions - only one of which is async - to extract the message from whatever type of input is provided, and then a separate - synchronous - function for converting that into an error. Feels clearer to me. What do you think?

PR #171, opened 2019-11-06
add traceId and silence config error log in release log handler

doramatadora:
## Why?

Make it easier to match up logs from Change API with logs from the release log handler.

## What has changed?

- Persist in the logger context `traceId` from the Change API message.
- Suppress logs when we fail to find `.github/runbooks.yml`

-   [ ] Bug fix
-   [ ] Feature
-   [ ] Breaking change
-   [ ] Review not required

## Reviewer checklist


lambdas/ingester/src/releaseLogHandler.js
```diff
@@ -212,9 +215,10 @@ const fetchRunbook = async (
 		commit,
 		systemCode,
 		gitRepositoryName,
-		githubData: { htmlUrl: gitRefUrl },
-		user: { githubName },
+		githubData: { htmlUrl: gitRefUrl } = {},
+		user: { githubName } = {},
 		eventID,
+		traceId: changeApiTraceId,
```
nl-ria: can we call the logging variable `traceId` instead of `changeApiTraceId`?

 This would greatly simplify the Splunk query and we can simply have a Splunk search like - 

`index="aws_cloudwatch" source="/aws/lambda/change-request-api*" OR source="/aws/lambda/biz-ops-runbook-md*" traceId="<uuid>"` which will showcase the entire user journey starting from Change API validation to every single step until and including the runbook.md pipeline

lambdas/ingester/src/releaseLogHandler.js
```diff
@@ -54,12 +55,14 @@ const createGithubAPIClient = (log = logger) => async (
 
 	if (!response.ok) {
 		requestOptions.headers.Authorization = `token: <redacted>`;
-		log.error({
-			event: 'GITHUB_API_FAILURE',
-			url,
-			requestOptions,
-			statusCode: response.status,
-		});
+		if (!silentMode) {
```
lucas42: Perhaps we could use a lower logging level (eg `logger.debug`) rather than wrapping the logging in extra logic?

PR #172, opened 2019-11-07
better logging and repository inference from release logs

doramatadora:
## Why?

High number of 404s from Github API and a high (70%) failure rate from the release log ingester.

## What has changed?

- we noticed we did not always have the owner `Financial-Times` in the inferred repository name from the Change API message
- we now check that the inferred repository name contains the repo owner; if it does not, prepend `Financial-Times/`
- We traced the 404s not related to missing owner to a non-compliant token. This token did not have read access to all repos. It's impossible to trace where the token came from. We are now using the same github token as Change API.

How the repo name is inferred from the Change API message payload:
1. `gitRepositoryName`, if it exists, otherwise
2. parse from `githubData.htmlUrl`, if it exists
3. bail, otherwise

It would be great to have a platform wide Github token to avoid this situation in the future.

-   [x] Bug fix
-   [ ] Feature
-   [ ] Breaking change
-   [ ] Review not required

## Reviewer checklist


lambdas/ingester/src/releaseLogHandler.js
```diff
@@ -15,20 +15,26 @@ const bizOpsApiKey = process.env.BIZ_OPS_API_KEY;
...
 		log.info(
-			{ event: 'BAIL_INSUFFICIENT_DATA', record },
+			{ event: 'BAIL_INSUFFICIENT_DATA' },
 			'Record did not contain commit',
 		);
 		return false;
 	}
 
-	return !!record.isProdEnv;
+	return !!isProdEnv;
```
nl-ria: think it would be a good idea to bail if `systemIsInBizOps` is false here too.

doramatadora: that's covered further down the line – we'd like to still expose logs for that situation

PR #182, opened 2019-11-21
refactor release handler

doramatadora:
## Why?

The `releaseLogHandler` has been giving us a lot of pain, and has reached unacceptable complexity.

## What has changed?

- removed all traces of `releaseLogHandler`
- added `releaseHandler`, triggered by the Change-API Kinesis stream; explanations below
- refactored the webhook to add runbook SHAs to every commit level entry in DynamoDB
- throttled requests to Biz-Ops API in the webhook
- throttled requests to DynamoDB to stick within read capacity bounds
- added autoscaling to DynamoDB, 1-25 read and write, aiming for 70% usage

Rather than relying on complex logic to ingest runbooks modified as a consequence of a commit (this is what `releaseLogHandler` used to do), `releaseHandler` looks for the results of ingest / validation in DynamoDb – and updates Biz-Ops with data from _all valid runbooks at that point in time_. Far better!

Taking the GitHub app out of beta (i.e. failing builds for invalid runbooks) should mean that there are never race conditions – by the time `releaseHandler` consumes a Change API message, we will have validated the runbook data.

-   [ ] Bug fix
-   [ ] Feature
-   [ ] Breaking change
-   [ ] Review not required

## Reviewer checklist


Makefile
```diff
@@ -101,9 +101,9 @@ move-asset-manifest:
 create-database:
 	aws cloudformation create-stack \
 	--region eu-west-1 \
-	--stack-name biz-ops-runbooks-md-data \
+	--stack-name biz-ops-runbook-md-data \
 	--template-body file://$$(pwd)/cloudformation/dynamodb.yaml \
-	--tags Key=description,Value="Data store for pull request evaulations of runbook.md files in repos" \
-	Key=systemCode,Value=yakfly \
+	--tags Key=description,Value="Data store for pull request evaluations of runbook.md files in repos" \
+	Key=systemCode,Value=biz-ops-runbook-md \
```
wheresrhys: 😂

Makefile
```diff
@@ -101,9 +101,9 @@ move-asset-manifest:
 create-database:
 	aws cloudformation create-stack \
 	--region eu-west-1 \
-	--stack-name biz-ops-runbooks-md-data \
+	--stack-name biz-ops-runbook-md-data \
```
wheresrhys: agree this is an annoying misnaming, but unless we plan to tear it down and recreate, probably shouldn't rename here

lambdas/ingester/src/commands/ingest.js
```diff
@@ -1,100 +1,61 @@
 const stripHtmlComments = require('strip-html-comments');
-const runbookMd = require('../lib/parser');
+const RunbookMd = require('../lib/parser');
 const { validate, updateBizOps } = require('../lib/external-apis');
-const {
-	updateSystemRepository,
-	systemHeadRequest,
```
wheresrhys: will wait for you to come in to talk through the new implementation, but one question I have is to make sure this checking for system existing in Biz Ops  happens somewhere in the github app flow

lambdas/ingester/src/commands/ingest/details.js
```diff
@@ -0,0 +1,13 @@
+const transformSOSResult = ({
```
wheresrhys: rename this file?

lambdas/ingester/src/commands/ingest/error-messages.json
```diff
@@ -0,0 +1,11 @@
+{
```
wheresrhys: couple of thoughts:
- more usual to use CONSTANT_CASE for keys of things like this
- Could also go with error codes - i.e. give each one a unique 3 digit number

serverless.yml
```diff
@@ -118,6 +118,17 @@ functions:
           arn: arn:aws:kinesis:eu-west-1:${env:AWS_ACCOUNT_ID}:stream/change-request-api-${env:ENVIRONMENT}-enriched-stream
           startingPosition: TRIM_HORIZON
           enabled: true
+  release:
```
wheresrhys: Why this as well as releaseLog?

Makefile
doramatadora: I won't tear down the stack. We'll just create a new one.

lambdas/ingester/src/commands/ingest.js
doramatadora: That check is triggered by all `ingest` calls under 3 conditions:
1. runbook is valid (structurally - no parse errors)
2. we have a non-empty system code
3. we are hoping to write to Biz-Ops

The app flow uses `ingest`, like all other flows.

serverless.yml
doramatadora: I'll just get rid of it. I can compare behaviours via the logs.

lambdas/ingester/src/commands/ingest/error-messages.json
doramatadora: One for another tidy-up PR?

PR #200, opened 2020-01-16
fix exporter generation

wheresrhys:
## Why?

Is broken
## What has changed?

Fixed reference to exported excluded field list


lambdas/ingester/src/lib/generate-from-biz-ops.js
```diff
@@ -59,15 +58,14 @@ exports.generate = async systemCode => {
 		)
 		.filter(({ type }) => !isForbiddenType(type))
 		.filter(({ name }) =>
-			[...systemSchema.minimumViableRecord]
+			[...(systemSchema.minimumViableRecord || [])]
```
wheresrhys:
```suggestion
			[...systemSchema.minimumViableRecord]
```

PR #203, opened 2020-02-07
Write runbook md url

wheresrhys:
## Why?

In the Biz Ops UI all the edit links point to the Biz Ops edit page even when the record is maintained in a runbook.md. There is no easy journey to finding the runbook.md file.

## What has changed?

This builds a url to the RUNBOOK.md file on master and adds it to the payload that is written to Biz Ops. Also a few minor lint and dependency changes (not deliberate, but should be fairly inert)


nl-ria:
⭐️

PR #204, opened 2020-02-12
Avoid insisting that users add disallowed fields

wheresrhys:
## Why?

Nikita observed that if she didn't include a repository field in her runbook.md she woudl be warned that she needed to specify a repository. However, repository is not allowed in RUNBOOK.md (connections should be added by other means), and the same is true of other blacklisted fields, so if the details are added to runbook.md, then on submission there is a parse error

## What has changed?
 For blacklisted fields, adds a note that the correct place to address the issue is in the biz ops website


lambdas/ingester/src/templates/components/output-fields.jsx
```diff
@@ -72,13 +73,20 @@ exports.ValidationErrors = ({ errors }) => {
...
-			facet,
-			{
-				props: { className: 'with-line-breaks' },
-				value: messages,
-			},
-		]),
+		rows: Object.entries(errors).map(([facet, messages]) => {
+			if (excludedProperties.includes(facet)) {
+				messages.push(
+					"Note that, in order to prevent this system's runbook blocking edits by other systems, this must be fixed by editing in https://biz-ops.in.ft.com, not by adding a field to your RUNBOOK.md.",
```
nl-ria: ermm, not sure I understand the explanation as to why we aren't allowed to add `repositories` field ? I thought its because of the BizOps GH importer controlling which value is added to this field.

Maybe make the text clearer ? something like - “no action required. This field needs to be manually added via BizOps UI because its controlled by other apps?”

PR #208, opened 2020-02-19
Add relationship editing step to quickstart

wheresrhys:
I just thought it's  anice tip to only generate the runbook.md _after_ editing relationships, because then you get to use the autocomplete


wheresrhys:
The main motivator for writing the quickstart is for existing projects (only 176 projects use runbook.md so far). Would like to try to drive more existing projects to use it, so getting the documentation in place for them

PR #210, opened 2020-03-05
upgrade tc libraries to 0.4.0

kiyaGu:
## Why?

We released version 0.4.0 of treecreeper which adds multiple choice and rich-relationship

## What has changed?

upgraded all tc libraries used by biz-ops-api to that version

None


wheresrhys:
let me know if you need help fixing tests. Or want me to take over this PR - The markdown parser code used here is very differenth to what you've worked on, so might not be very easy to figure out what's going wrong

wheresrhys:
Have tested the UI and it now all works

PR #218, opened 2020-04-01
Functional refactor 1

wheresrhys:
## Why?

RUNBOOK.md is very difficult to support, and already feels like a legacy system. It was broken for almost 2 whole days this week.

I've identified 4 main problems
1. The various stages of ingestion (find runbooks, parse, score, ...) are not organised like a pipeline of actions, which makes things a bit incomprehensible. The class-based codebase also makes it different to work out what persistent state exists at the point a method is called.
2. It's very difficult to debug from the logs, or even to be sure if the app is broken
3. The only way to be confident a change works is to deploy to prod and try on a few repos in different scenarios
4. Nothing alerts when it is broken

## What has changed?

I've made a start on 1. Previously the ingest function a) contained parse, score and ingest functionality b) it was possible to parse and score but skip ingest and vice versa. This was crying out to be split into two functions.

I've also moved the stash in dynamoDB functionality out of the class inheritance structure, where functions were split across multiple subclasses, and into a single module which explicitly has the runbooks data passed in (that could've been a separate PR, but what they hey)

The plan is to gradually address points 1 to 4 over a series of PRs so the app is more supportable.


lucas42:
Rhys, did you discuss your approach with anyone before you set on this refactor?  This is a very big changeset and I don't think any of us have enough context to give it a thorough review.

Perhaps consider breaking it into smaller chunks?  Or we could have a discussion about it to give people the necessary context?

katsuraku:
I haven't given it a thorough review - i'm pretty/very hazy on the runbook.md - but it looks more readable now. I appreciate this use of modules with readable params over class methods for sure

wheresrhys:
> This is a very big changeset and I don't think any of us have enough context to give it a thorough review.

I agree with not having enough context. This PR could be split in 2, but even then I doubt anyone would have enough context. And the vast majority of the PR is to do with one of the changes. This app is a monster, and it's not going to be pretty unpicking it.

Very happy to run something for the team discussing this, although I am finding I learn as I go along so, beyond the changes I've already made, Any information I can give will be very broad strokes

wheresrhys:
Discussed in show & tell.

I'll add the hand drawn diagrams I have which describe this refactoring journey. Came to an agreement that investing time in getting understanding of the detail of changes I'm making is premature at this stage, when it's mainly/entirely moving existing code around into a more comprehensible structure that resembles the actual flow that it implements. Once we're there, and looking at improving supportability, that would be a better time for others to invest their time and energy. 

Sanity checking reviews of PRs are still welcome though, and I will continue to try to keep them small (although many will still look big going by line count as that is, unfortunately, what you get when restructuring code, even if the behaviour is not changing)

PR #225, opened 2020-04-09
Functional3

wheresrhys:
## Why?

Part 3 of the move to a more functional approach. 

## What?
This one moves the code that puts a status on github PRs into a standalone module.

The change, once again, looks bigger than it is as runbook-md-app has been deleted and runbook-processor has been renamed to runbook-md-app, so the diff is huge.

Other significant changes:
- The checkSystemCodeExists function now returns true or false, with only 500 errors remaining unhandled. The fancy logic around 404's that was there made the code less reusable and didn't actually help in debugging anyway.
- generation of messages for the github status is a little neater I hope. I've removed some logic which was undocumented and added during a big bang release ages ago. It handles _some_ sort of edge case on repos with multiple runbooks. I'm sure we'll be told if things break, and then we can reimplement and document properly if it's actually still needed
- The things that are extracted out of the class-based app are now no longer passed the app as a parameter. Before this stage the following variables are created from the app in order to be passed around:

```js
	const webhookId = { repository: app.repository, commitSha: app.sha };
	const { logger } = app;
	const { runbooks } = app;
```

Still not the finished article, but does help with reducing the lack of clarity about which bits of the app are used where, so we can continue to chip away at it more easily.


lambdas/ingester/src/commands/ingest.js
```diff
@@ -26,7 +26,14 @@ const ingest = async payload => {
 
 	// 3. check if the system code exists in Biz Ops
 	// to avoid creating new systems
-	await checkSystemCodeExists(systemCode, details);
+	const systemCodeExists = await checkSystemCodeExists(systemCode);
+
+	if (!systemCodeExists) {
```
wheresrhys, line 31: This used to be in checkSystemCodeExists. Moved here to improve reusability of that utitlity

